PR #261, opened 2025-03-11
Yes you can open pull requests now

sp2xdev:
Thanks.


src/spice2x/.gitignore
```diff
@@ -1,4 +1,5 @@
 .vscode/*
+docs/*
```
sp2xdev, line 2: :octocat:

Issue #276, opened 2025-03-26
Image scaling breaks if base resolution is too high, or if in portrait orientation


bicarus-dev:
## Describe the issue
When using -forceres with high values, say, 2560x1440. With image scaling, if you zoom out far enough, eventually scaling will break because we run out of texture to render.

At 4K, things break almost immediately even at 1.0 scaling.

Additionally, when in portrait orientation, scaling factor seems to be broken in weird ways.

PR #277, opened 2025-03-26
sdvx: landscape mode

bicarus-dev:
n/a

## Description of change
Add `SDVX Landscape Mode` option, allowing you to play SDVX in landscape monitor orientation, with black borders on left and right - similar to how EAC does it. This is meant to be an alternative to what people do today, which is launching in windowed mode.

By default, with just `-sdvxlandscape` enabled, EG will launch at 1920x1080 (transposed from the original 1080x1920 resolution).

It is possible to combine with `-forceres` to launch at higher resolutions. At some point though, when rendering at large resolutions past 1440p, at some point scaling may break due to #276.

Disabled on 32-bit SDVX since dx9 `CreateDeviceEx` call is required; `CreateDevice` won't work. For the same reason this isn't being made a generic option for museca/rb/etc. In theory you can supply `-sdvxlandscape` to games like IIDX and play landscape games in portrait mode, although again, scaling my break due to #276.

Also, add new options (bottomleft/right) for subscreen overlay position so that they don't overlap with the main screen in landscape mode.

## Compiling
🙂

## Testing
Tested a couple versions of VW and EG.

### DIFF
--- a/src/spice2x/launcher/options.cpp
+++ b/src/spice2x/launcher/options.cpp
@@ -212,7 +212,7 @@ static const std::vector<OptionDefinition> OPTION_DEFINITIONS = {
         .desc =
             "For full screen mode, forcibly set a custom resolution.\n\n"
             "Works great for some games, but can COMPLETELY BREAK other games - YMMV!\n\n"
-            "If you are using -landscape, put the TARGET monitor resolution in this field.\n\n"
+            "If you are using -sdvxlandscape, put the TARGET monitor resolution in this field.\n\n"
             "This should only be used as last resort if your GPU/monitor can't display the resolution required by the game",
         .type = OptionType::Text,
         .setting_name = "1280,720",


src/spice2x/hooks/graphics/backends/d3d9/d3d9_device.cpp
```diff
@@ -718,7 +718,37 @@ void SurfaceHook(IDirect3DDevice9 *pReal) {
...
+            topSurface, &originRect,
+            D3DTEXF_LINEAR);
+        if (hr != D3D_OK) {
+            log_misc("graphics::d3d9", "StretchRect backbuffer failed (forced landscape)");
+        }
+    } else {
+        // stretch to add top/left offset to avoid going negative
+        hr = pReal->StretchRect(
+            backbuffer, nullptr,
+            topSurface, nullptr,
```
bicarus-dev, line 745: `nullptr` -> `&targetRect` (spotted after merge)

Issue #292, opened 2025-04-05
Build system improvements and suggestions


ASleepyCat:
A few ideas to improve the current build system:

* Use Github Actions for CI
  * Eliminates need to trust people that their code compiles when instead you can set up requirements to have CI passing before being allowed to merge pull requests
  * Also eliminates the need to compile locally to test out changes since you can download the build artifact
* Use `ninja` to compile
  * Improves build times by a small bit. On my system with 32GB of RAM and a 7800X3D, switching to `ninja` reduces compile times by 13 seconds (from 2m43.853s with `make`).
  * Only requires a few changes to the `spicetools/deps` Dockerfile and `build_all.sh`
* Use vcpkg for external dependencies
  * Reduces the amount of dependencies stored in the repo
  * Does require more dev setup though, and not all packages are in vcpkg (e.g. `headsocket`)
  * List of packages in vcpkg:
    * `asio`
    * `discord-rpc`
    * `cpu_features`
    * `fmt`
    * `hash-library`
    * `http-parser`
    * `imgui`
    * `linalg`
    * `minhook`
    * `rapidjson`
    * `readwriterqueue`
    * `robin-hood-hashing`
    * `stackwalker`
    * `stb_image_write` via `stb`
    * `tinyxml2`

ASleepyCat:
> * I don't know about switching away from CMake, at least for the primary build system. If you want to submit build scripts that can be used for dev builds using ninja, that's totally OK with me.

Just to clarify, `ninja` would replace `make`. We'd still use CMake to generate the `ninja` build files.

> * For external dependencies, there are two issues:
> 
> 
> 1. It turns out some of them have heavily modified versions. ImGui is one of them - the DX9 renderer is a custom one based on a version from 6 years ago, for example. Updating to the latest one breaks in very subtle ways. So someone needs to do the work of finding out the right version and ensuring that there is no custom modifications.

vcpkg has a way to supply patch files for packages, so theoretically we could keep a patch file that contains all our spice-specific changes.

> 2. We rarely ever update external dependencies, usually out of fear that something will break, or just usually not worth the effort. The concern here is that because spice supports huge variety of hardware and game versions, it's impossible to test everything, and updating external libraries can easily break really obscure uses cases, which we want to avoid.
> 
> 
[...]

PR #293, opened 2025-04-06
Build system improvements

ASleepyCat:
#292

## Description of change
This PR adds CI with Github Actions. Runs on every push and pull request. If a new commit is pushed, any previous runs are cancelled for that branch.

In total, these optimisations (on a 7800X3D with 32GB 6000MT/s RAM and an NVMe Gen4 SSD):
* Reduces `spicetools/deps` image size by around 200 MB, and build time by almost 50%
* Reduces compilation time by around 9%

Changes:
* Pull archlinux image from GHCR so that we don't run into pull limits from docker.io
* Use `base-devel` archlinux image
  * Reduces image build time from 141 seconds to 109 seconds on my machine
  * Clean up unneeded packages
    * Small reduction in `spicetools/deps` image size (4.35 GB vs 4.14 GB)
* Use precompiled `yay` binary
  * Reduces image build time from 163 seconds to 141 seconds on my machine
* Use `nproc` to get the number of cores instead of parsing `/proc/cpuinfo` with `awk`
* Use `ninja` to compile
  * Speeds up compile times a little, around 13 seconds on my machine
* Don't run `docker run` in interactive TTY for the bash script
  * Github Actions disallows interactive TTYs
  * Still allowed for the batch script
* `time` the compilation process
* Enable BuildKit in the bash script
* Don't use `USER` instruction
  * Breaks Github Actions, doesn't seem to affect build?
  * More info: https://docs.github.com/en/actions/sharing-automations/creating-actions/dockerfile-support-for-github-actions#user
* Add executable bits to bash scripts

## Compiling
Check CI status.

## Testing
Tested with docker on Windows.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
...
+      run:
+        working-directory: ./src/spice2x
+    steps:
+      - uses: actions/checkout@v4
+      - name: Compile
+        run: ./build_docker.sh
+      - uses: actions/upload-artifact@v4
+        with:
+          name: spice2x
+          path: src/spice2x/bin
```
sp2xdev, line 21: It's not a huge deal, but could these just grab the zip file from `/dist`?

src/spice2x/build_docker.sh
```diff
@@ -1,6 +1,4 @@
 #!/bin/bash
-export DOCKER_BUILDKIT=0
-
 docker build --pull $PWD/external/docker -t spicetools/deps --platform linux/x86_64
 docker build --build-context gitroot=$PWD/../../.git . -t spicetools/spice:latest --no-cache
-docker run --rm -it -v $PWD/dist:/src/src/spice2x/dist -v $PWD/bin:/src/src/spice2x/bin spicetools/spice
+docker run --rm -v $PWD/dist:/src/src/spice2x/dist -v $PWD/bin:/src/src/spice2x/bin spicetools/spice
```
sp2xdev, line 4: should update the .bat as well (don't worry too much about testing it, if you don't run Windows)

ASleepyCat: I didn't update the batch script since it prevents you from Ctrl+C'ing the build to cancel it (without also stopping the container in docker). I only made the change to the bash script for Github Actions support.

.github/workflows/ci.yml
ASleepyCat: Unfortunately there's no way to directly upload a file as an artifact, they have to be stored inside a zip file. So the archive structure would look like this:

```
spice2x.zip
├───spice2x-x-y-z.zip
│   ├───spice2x
│       ├─── ...
```

~~The `dist` archives also contain the source code which I don't think really makes sense for an artifact.~~ Nevermind, the source code is part of the artifacts. I guess the only issue would be the somewhat cumbersome archive structure.

sp2xdev: OK, no worries then.

src/spice2x/build_docker.sh
sp2xdev: I see, thanks.

PR #294, opened 2025-04-06
overlay: update UI for card manager

bicarus-dev:
n/a

## Description of change
Change the usage pattern for Card Manager UI.

### Previous behavior

Previously, it was a list of cards that you select, and click a button to insert as P1 or P2.

This is fine, but users got confused when they pressed `Insert Card` overlay key and got a different card inserted, or when auto-insert didn't work as expected.

Additionally, if you play a game without continue (IIDX/DDR for example), then it's cumbersome to bring up the overlay and click on insert card every time.

### New behavior

Insert Card overlay now lets you pick a card and *slot* it into P1 or P2 side.

From that point on, the slotted card number is used as the override, replacing what was previously passed to `-card0` or `-card1`. This means that `Insert Card` button and auto-insert will pick up the new card and use that instead.

## Compiling
👍 

## Testing
Tested 1p-only game (popn) and 2p games (DDR, IIDX).

### DIFF
--- a/src/spice2x/overlay/windows/card_manager.cpp
+++ b/src/spice2x/overlay/windows/card_manager.cpp
@@ -91,6 +91,10 @@ namespace overlay::windows {
             }
             ImGui::EndTable();
         }
+
+        ImGui::Spacing();
+        ImGui::Spacing();
+
         ImGui::TextColored(ImVec4(1, 0.7f, 0, 1), "Available cards");
         build_card_list();
         ImGui::Separator();


src/spice2x/misc/eamuse.h
```diff
@@ -66,6 +68,7 @@ void eamuse_set_keypad_overrides(size_t unit, uint16_t keypad_state);
 void eamuse_set_keypad_overrides_bt5(size_t unit, uint16_t keypad_state);
 void eamuse_set_keypad_overrides_reader(size_t unit, uint16_t keypad_state);
 void eamuse_set_keypad_overrides_overlay(size_t unit, uint16_t keypad_state);
+void eamuse_set_keypad_overrides_card_manager(size_t unit, uint16_t keypad_state);
```
sp2xdev: Unused.

PR #299, opened 2025-04-17
Fix HID string ranges not being retrieved correctly

ASleepyCat:
## Description of change
Currently, retrieving HID strings using from string index minimum/maximum ranges is broken as it uses the same index for each usage cap.

Now it properly ranges over the minimum/maximum ranges.

## Compiling
Check CI.

## Testing
Tested with Beef Board firmware using string index ranges in the USB HID descriptor.


ASleepyCat:
> Would this be broken as well for value output caps?
> 
> line 662
> 
> ```c++
> // get string index
> ULONG string_index = 0;
> if (value_caps.IsStringRange && value_caps.Range.StringMin != 0) {
>     string_index = value_caps.Range.StringMin
>             + static_cast<ULONG>(value_output_caps_list.size()) - 1;
> } else if (!value_caps.IsStringRange && value_caps.NotRange.StringIndex != 0) {
>     string_index = value_caps.NotRange.StringIndex;
> }
> ```

For value caps they work as expected. They get pushed to `value_output_caps_list` before retrieving the string.

ASleepyCat:
Looks like whatever HID parser that Infinitas is using breaks with ranged string indexes anyway, so I can't even use it in my firmware lol

Issue #41, opened 2023-07-31
output taped led colors through spiceapi


hlcmOxO:
spiceapi now can only support outputting the average of taped led colors(Left Wing Avg R in sdvx), is there a way to make it output the color for each individual LED instead of the average value?

ASleepyCat:
How should this be accessible to the user? I highly doubt we'd want to clog the lights menu with glizzilion entries, should we hide from the the GUI and only expose them via spice API? Maybe make a UI where the tape LEDs can be configured to either be averaged or individually assignable?

PR #315, opened 2025-05-05
Add support for outputting tape LED data for IIDX via spiceapi

ASleepyCat:
#41 (partially)

## Description of change
This PR adds support for outputting IIDX tape LED data via spiceapi. It returns the data as a `dict`, with the key being the LED name and the value being the list of RGB values.

An example script using the new API is available in this Beef Board PR: TBD.

Demo use case: TBD

## Testing
Tested in IIDX 32 with custom Beef Board firmware along with a Python script to read and send the tape LED data over USB.


src/spice2x/api/modules/iidx.cpp
```diff
@@ -69,4 +70,44 @@ namespace api::modules {
...
+            "side_panel_left_outer",
+            "side_panel_left",
+            "side_panel_right_outer",
+            "side_panel_right_inner",
+            "side_panel_right"
+        };
+
+        Value response_object(kObjectType);
+
+        // Iterate through each device and dump its lights data into the response
```
sp2xdev: Suggestion: allow the request to optionally specify which light(s) to get, instead of all the lights.

ASleepyCat: Done.

Issue #316, opened 2025-05-06
Support LovePlus Arcade Camera


Cetaceaqua:
I know there is very little hardware info about the cab, and this game is rarely played.

Is there any chance that its camera support can be added? So we can try the collab function of scanning QR codes with Nintendo 3DS New LovePlus game.

Konami promo pop introducing the collab function: [Wayback Machine](https://web.archive.org/web/20150506224248/http://www.konami.jp/am/ac_ope/pdf/20120213_01.pdf)

I notice that there is `-noCamera` param in `/src/spice2x/games/loveplus/loveplus.cpp` `GetCommandLineA_hook()`, is this used to temporarily block the camera?

Just inquiry, thanks for your attention.

PR #318, opened 2025-05-07
loveplus: enable cam access

bicarus-dev:
#316 
(this does not fix the issue!)

## Description of change
Add an option to omit `-noCamera` option. Doing so allows the game to boot and search for a compatible camera using KS API, but will likely fail since we don't know what cameras are compatible. You end up with a camera error during boot, which can be ignored by flipping a switch in the test menu to disable the camera.

No hooks are currently done to make the camera work!

## Testing
Tested to see that turning this on triggers the camera error. No known camera at the moment.

### DIFF
--- a/src/spice2x/games/loveplus/loveplus.cpp
+++ b/src/spice2x/games/loveplus/loveplus.cpp
@@ -119,11 +119,13 @@ namespace games::loveplus {
     }
 
     static LPSTR __stdcall GetCommandLineA_hook() {
-        std::string lp_args = "-win -noWatchDog -noIOError -noIrda -notarget";
-        if (!CAMERA_ENABLE) {
-            lp_args += " -noCamera";
+        static std::string lp_args = "-win -noWatchDog -noIOError -noIrda -notarget";
+        static std::string lp_args_nocamera = lp_args + " -noCamera";
+        if (CAMERA_ENABLE) {
+            return lp_args.data();
+        } else {
+            return lp_args_nocamera.data();
         }
-        return lp_args.data();
     }
 
     LovePlusGame::LovePlusGame() : Game("LovePlus") {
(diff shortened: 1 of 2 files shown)


src/spice2x/games/loveplus/loveplus.cpp
```diff
@@ -117,8 +119,10 @@ namespace games::loveplus {
     }
 
     static LPSTR __stdcall GetCommandLineA_hook() {
-        static std::string lp_args = "-win -noCamera -noWatchDog -noIOError -noIrda -notarget";
-
+        std::string lp_args = "-win -noWatchDog -noIOError -noIrda -notarget";
```
sp2xdev: still needs `static` keyword

Cetaceaqua:
Hi!

Thanks for the fix, I have tried the latest Github Actions build.

Fortunately, it recognized my laptop camera directly, and the qr code scanning worked fine. In addition to the fact that focusing is very painful, the arcade cab should use a fixed-focus camera.

According to Game Saru's [video](https://www.youtube.com/watch?v=J2VVmenqcN0&t=457s), the cab camera is portrait mounted, which is why my camera is rotated 90°.

Here is my laptop built in camera info:

```
{"Webcam Name:":"Integrated Camera (5986:115f)","Frame rate:":"30 FPS","Webcam Resolution:":"1280x720","Aspect Ratio:":"1.78","Webcam MegaPixels:":"0.92 MP","Camera Facing:":"environment","Built-in Microphone:":"None","Built-in Speaker:":"None","PNG File Size:":"257.45 kB","JPEG File Size:":"29.89 kB","Number of Colors:":"35281","Brightness:":"27.69 %","Hue:":"6.36 %","Saturation:":"43 °","—":"—"}
```

PR #327, opened 2025-05-14
launcher: option to make command line args take precedence

### LINKED ISSUE #190: Command line option should have precedence
As title. When specifying an option through command line, it should overwrite existing configuration specified through spicecfg. 

Likewise, if the same option is specified multiple times in command line, the last occurrence should take effect, although this is an edge-case.

This would speed up development and come in handy for users who connect to multiple servers.

bicarus-dev:
Fixes #190 

## Description of change
When `-cmdoverride` is specified, any command line arguments are taken first before merging with what is in the config file.

## Testing
Still testing...

### DIFF
--- a/src/spice2x/launcher/options.cpp
+++ b/src/spice2x/launcher/options.cpp
@@ -44,7 +44,6 @@ static const std::vector<std::string> CATEGORY_ORDER_NONE = {
 };
 
 bool launcher::USE_CMD_OVERRIDE = false;
-bool launcher::LOG_CMD_OVERRIDE_HELP = false;
 
 /*
  * Option Definitions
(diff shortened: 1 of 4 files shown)


src/spice2x/launcher/options.cpp
```diff
@@ -2301,6 +2313,16 @@ static const std::vector<OptionDefinition> OPTION_DEFINITIONS = {
         .game_name = "LovePlus",
         .category = "Game Options (Advanced)",
     },
+    {
+        // OptionConflictResolution
+        .title = "Command Line Args Overridee",
+        .name = "cmdoverride",
+        .desc = "By default, option values in spicecfg take precedence over comamnd-line args, for legacy compat.\n"
```
LupinThidr: typo (comamnd + overridee)

Issue #373, opened 2025-09-19
Use format from C++20 stl instead of fmt?


GEEKiDoS:
The fmt library produces tons of warnings when compiling with MSVC, Can we get rid of it?

bicarus-dev:
Revisiting this - the biggest blocker would be `filesystem::path` handling which is included in C++ 26. It'll be a while until toolchains support it. We also adopted a separate XP toolchain which is pinned to a certain version of LLVM so that might be a challenge.

PR #374, opened 2025-09-19
Fix msvc compiler warnings

GEEKiDoS:
To be noticed I edited thirdparty header for fmt library (#373), idk if it is acceptable

### DIFF
--- a/src/spice2x/.gitignore
+++ b/src/spice2x/.gitignore
@@ -24,5 +24,3 @@ external/cv2pdb/*
 .vs
 out
 CMakeSettings.json
-
-RCa22448


GEEKiDoS:
I have no idea what is that file tbh

src/spice2x/.gitignore
```diff
@@ -24,3 +24,5 @@ external/cv2pdb/*
 .vs
 out
 CMakeSettings.json
+
+RCa22448
```
sp2xdev: Can you remove this - don't think this is expected from MSVC

PR #381, opened 2025-09-26
iidx: apply signature patch for ASIO compatibility fix in iidx32

bicarus-dev:
n/a

## Description of change
Apply this fix automatically when needed - https://github.com/two-torial/sp2xpatcher/pull/33

## Testing
Tested iidx31 to see that the patch does not apply
Checked -audiohookdisable to see that the logic gets skipped
Tested 2025041500 and 2025082500.


bicarus-dev:
I don't have new data so I can't test

PR #398, opened 2025-10-09
Fixed an issue where ext (release_code) wasn't detected and replaced correctly when using user specified bootstrap.xml path.

AllanCat:
#0

## Description of change
The original launcher::detect_bootstrap_release_code() wasn't detecting user specified bootstrap.xml path, and was causing issues for patch_manager and other ext check spots prior to ea3 init (as ea3 does proper path check itself).
I've added the user path check and proper xml parsing from detect_gameversion().

## Testing
Tested with prop/bootstrap.xml presents.
Tested with prop/bootstrap.xml not presents and -b /path/to/bootstrap.xml

### DIFF
--- a/src/spice2x/launcher/options.cpp
+++ b/src/spice2x/launcher/options.cpp
@@ -2626,41 +2626,7 @@ std::string launcher::detect_bootstrap_release_code(const std::string& bootstrap
     tinyxml2::XMLDocument bootstrap;
     if (bootstrap.LoadFileA(bootstrap_path.c_str()) != tinyxml2::XML_SUCCESS) {
         log_warning("options", "unable to parse {}", bootstrap_path);
-        log_warning("options", "trying again with improper XML header removed...");
-
-        // below: dirty hack to deal with bad ea3 xml file with comments in the beginning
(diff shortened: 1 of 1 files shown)


src/spice2x/launcher/options.cpp
```diff
@@ -2608,14 +2608,59 @@ std::vector<Option> launcher::merge_options(
...
+    }
 
     // load XML
     tinyxml2::XMLDocument bootstrap;
     if (bootstrap.LoadFileA(bootstrap_path.c_str()) != tinyxml2::XML_SUCCESS) {
         log_warning("options", "unable to parse {}", bootstrap_path);
-        return "";
+        log_warning("options", "trying again with improper XML header removed...");
+
+        // below: dirty hack to deal with bad ea3 xml file with comments in the beginning
```
sp2xdev: Do we need this?

AllanCat: This section was directly ported from ea3 parsing, I'm not sure if bootstrap will encounter same issue as ea3-config but just as a failsafe method.

AllanCat: You're right. After more testings on bootstrap.xml, I found tinyxml2 was already strong enough to parse various xml headers (even worked with garbled format).

PR #400, opened 2025-10-12
add option to remove winscard dependency

kazakhstan-sila:
n/a

## Description of change
Add CMake option to disable SCard functionality. Needed by some builds of WINE, where distributor opted to not enable scard functionality for whatever reason

## Testing
Building with `cmake -DWITH_SCARD=OFF` does not link against winscard
Building without this option at all links against winscard


src/spice2x/launcher/options.h
```diff
@@ -157,10 +157,12 @@ namespace launcher {
             CardIOHIDReaderSupport,
             CardIOHIDReaderOrderFlip,
             CardIOHIDReaderOrderToggle,
+#ifdef WITH_SCARD
```
sp2xdev: remove the ifdef around this, we don't want to have different enum values across build variants

sp2xdev: instead set disabled and hidden properties to true if missing scard dependency in options.cpp

Issue #150, opened 2024-03-17
DDR: Fix foot detection on 64-bit p4io


TheDDREnjoyer:
## Description of change
My previous patch that added p4io support hardcoded the timer source that the game was using to determine tick time: `timeGetTime()`.  This patch swaps that call out with a shim that will attempt to get the timestamp from `arkGetTickTime64()` in `ark*.dll`. If for whatever reason `arkGetTickTime64` is not present in the game dll, it will fall back to `timeGetTime`.

This patch maybe? also fixes step detection in bio2, if that was a problem before? I don't have data that works properly with it so I can't test it, but looking around in IDA it seems like it has the same issue as p4io (p3io inserts the required timestamp, p4io needs it to come from the buffer returned by libacio, and bio2 looks extremely similar). If anyone tries it out and it works let me know I'd be very interested to hear. Also, my email is in the patch file if you ever want to chat outside of GH.

## Testing
MDX was booted with 64-bit versions of `arkmdxp3.dll` and `arkmdxp4.dll` and the tutorial was played to ensure that steps could be detected in both architectures.

## Patch file
git format-patch
[0001-ddr-Fix-foot-detection-on-64-bit-p4io.patch](https://github.com/spice2x/spice2x.github.io/files/14625925/0001-ddr-Fix-foot-detection-on-64-bit-p4io.patch)

LupinThidr:
On bio2 the arrow recepticle(?) would move when arrows were stepped, but actual notes would just fall past as if nothing was being hit, is that what you were referring to by "foot detection"?
64-bit modules for A20+ (including arkmdxbio2.dll) are supposedly public now.

TheDDREnjoyer:
> On bio2 the arrow recepticle(?) would move when arrows were stepped, but actual notes would just fall past as if nothing was being hit, is that what you were referring to by "foot detection"? 64-bit modules for A20+ (including arkmdxbio2.dll) are supposedly public now.

yeah, that's exactly the same issue that p4io was having because of the missing timestamp

Issue #147, opened 2024-03-15
DDR: P4IO (white cab support)


TheDDREnjoyer:
## Description of change
Adds support to emulate the p4io board used in white DDR cabs

## Testing
MDX was booted with arkmdxp4.dll and the test mode was used to ensure lights and buttons were working, in addition, a few songs were played in the game.

## Patch file
git format-patch
[0001-ddr-p4io-support.patch](https://github.com/spice2x/spice2x.github.io/files/14609464/0001-ddr-p4io-support.patch)

LupinThidr:
Other than minor typos and variable name case that aren't really relevant, this looks good!
But shouldn't arkmdxp4 be under arkmdxp3.dll in launcher.cpp? That way it shouldn't break default behavior (p3io) or timing.

TheDDREnjoyer:
> Other than minor typos and nitpicks that aren't really relevant, this looks good! But shouldn't arkmdxp4 be under arkmdxp3.dll in launcher.cpp? That way it shouldn't break default behavior (p3io) or timing.

That seems reasonable to me, there was no rhyme or reason to why I made it prefer p4io

PR #439, opened 2025-12-09
MDX: Fixed Status Control Buffer Emulators (P4IO Timing Fix)

cchike:
## Description of change
In its current state, the emulated `libacio2.dll` functions `ac_io_mdxf_get_control_status_buffer` and `ac_io_mdxf_update_control_status_buffer` only operate using a single head pointer (`COUNTER`) for the ring buffers of `[panel_state,timestamp]` (simplified for brevity here) polls for both players, when they should be maintaining a separate head pointer for each ring buffer. On every frame, `arkmdxp4.dll` makes calls to `ac_io_mdxf_update_control_status_buffer`, which [in this emulated version] advances the head pointer, creates a new entry for the latest panel state info along with the current time, and stores it at the head of the ring buffer. This function is called for each player: After the call for 1P, the head pointer is advanced by 1, so by the time it's called for 2P, the head pointer had already been advanced by 1, meaning the new entry for 2P's panel state ring buffer is written two entries ahead of where the last entry for 2P was written. This is an example of how the entries were written for each ring buffer:

```
Frame 1:
1P[0] = t0
2P[1] = t0

Frame 2: 
1P[2] = t1
2P[3] = t1

Frame 3:
1P[4] = t2
2P[5] = t2

Frame 4:
1P[6] = t3
2P[0] = t3

...

Frame 7:
1P[5] = t6
2P[6] = t6

Frame 8:
1P[0] = t7
2P[1] = t7
```

This is an issue because `arkmdxp4.dll` calls `ac_io_mdxf_get_control_status_buffer` on every frame to load the last 7 polls for each player, which depends on the ring buffer entries being in reverse chronological order to properly determine when an arrow is pressed and when it was released, but this is what it sees from the ring buffers after 8 frames:

```
1P: t7 -> t3 -> t6 -> t2 -> t5 -> t1 -> t4
2P: t7 -> t3 -> t6 -> t2 -> t5 -> t1 -> t4
```

In this case, `arkmdxp4.dll` assigns step times based on polls that are out of order, resulting in unpredictable timestamp assignment to each step, with the issue compounding at lower framerates since the timestamp deltas between each poll will be greater (see side note). After implementing two separate head pointers for each ring buffer, those same two arrays become:

```
1P: t7 -> t6 -> t5 -> t4 -> t3 -> t2 -> t1
2P: t7 -> t6 -> t5 -> t4 -> t3 -> t2 -> t1
```

This is how `arkmdxp4.dll` expects this data to be structured, so timing should become more consistent after this change.

Side note: in this particular p4io implementation, the only place the polls are updated is once per frame on the calls to `ac_io_mdxf_update_control_status_buffer`, meaning the polling rate of the controller used is locked to the framerate of the game. There should be another thread not tied to the main one that listens to input from the OS and updates these ring buffers on a much faster cadence.

## Testing
I played a few songs to make sure nothing was broken. I also attached a debugger and verified the ring buffer entries are now being returned in reverse chronological order as intended.

### DIFF
--- a/src/spice2x/acio/mdxf/mdxf.cpp
+++ b/src/spice2x/acio/mdxf/mdxf.cpp
@@ -11,7 +11,8 @@
 const size_t STATUS_BUFFER_SIZE = 32;
 
 // static stuff
-static uint8_t COUNTER = 0;
+static uint8_t COUNTER_P1 = 0;
+static uint8_t COUNTER_P2 = 0;
 
 // buffers
 #pragma pack(push, 1)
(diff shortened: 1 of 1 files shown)


bicarus-dev:
Related: #147 #150

cchike:
> Is libacio2 in the PR body just a typo?

No, that’s the file that reads input from the pad and records it. I’ve verified on my white cab running official hardware with logging functions I patched into `libacio2.dll`. It’s also how I was able to verify the structure of the ring buffers.

PR #441, opened 2025-12-11
MDX: Fixed Ring Buffer Traversal Direction (P4IO Timing Fix) 

cchike:
## Description of change
In the real `libacio2.dll` implementation, `ac_io_mdxf_get_control_status_buffer` returns the ring buffer entry at `(head - index) mod size` but this function was using `(head + index) mod size` instead. Basically, `index` is supposed to indicate how many entries back, not forward. This was causing the polling history of each player to be returned backwards. A few other things were updated to bring the function's behavior closer to the real implementation:

1. Number of ring buffer entries per player were increased to 16 (this is the size of them in the real implementation)
2. `ac_io_mdxf_get_control_status_buffer` now returns the current head of the ring buffer (instead of a success/failure boolean), which is what `arkmdxp4.dll` expects.

## Testing
I played a few songs to make sure nothing was broken. I also attached a debugger and verified the ring buffer entries are *actually* now being returned in reverse chronological order as intended.


src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -41,30 +41,50 @@ static uint64_t arkGetTickTime64() {
...
-            // fill with zeros on unknown node
-            memset(buffer, 0, STATUS_BUFFER_SIZE);
-            return false;
-        }
+        // Select player-specific state
+		uint8_t head;
+		uint8_t (*buffer)[STATUS_BUFFER_SIZE];
+		size_t size;
+		
+		 if (node == 17 || node == 25) {
```
bicarus-dev: nit:

1. alignment for line 57
1. don't align equal signs in a column

src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -41,30 +41,50 @@ static uint64_t arkGetTickTime64() {
  * Implementations
  */
 
-static bool __cdecl ac_io_mdxf_get_control_status_buffer(int node, void *buffer, uint8_t a3, uint8_t a4) {
+static uint64_t __cdecl ac_io_mdxf_get_control_status_buffer(int node, void *out, uint8_t index, uint8_t head_in) {
 
+	// Default error value (matches original mask behavior)
+    auto error_ret = (uint64_t)(node - 0x11) & 0xFFFFFFFFFFFFFF00;
```
bicarus-dev: `static_cast`

src/spice2x/acio/mdxf/mdxf.cpp

src/spice2x/acio/mdxf/mdxf.cpp

PR #446, opened 2025-12-13
MDX: Decoupled Input Polling Rate From Game Refresh Rate (P4IO Timing Fix)

cchike:
## Context
In the real implementation of `ac_io_mdxf_update_control_status_buffer`, it doesn't actually add entries to the pad state ring buffers (it only marks the head entry as the "last consumed" entry), and `arkmdxp4.dll` expects the ring buffers to be updated at regular intervals inside `libacio2.dll` at ~250Hz per player. Instead, in the emulated version, the updates happen in this function and that regular interval is established internally using the game's refresh rate (60Hz, 120Hz, 144Hz, etc.). This effectively locks the polling rate of pad input to that refresh rate, which is quite low for pad input.

## Description of change
The polling rate lock was circumvented by calling `ac_io_mdxf_update_control_status_buffer` as soon as the game receives controller input in `rawinput.cpp`. A simple wrapper function `mdxf_poll` was exposed so press and release events could be added to the ring buffers in real time as they happen. After each controller event in `rawinput.cpp`, `mdxf_poll` calls `ac_io_mdxf_update_control_status_buffer` for each player, checking the pad's current state and adding it to the ring buffer if it's changed since the last call. This results in the 0 -> 1 (press) and 1 -> 0 (release) state transitions used for gameplay to be based only on the real-time events inserted from `rawinput.cpp`, while leaving the rest of the entries generated from the main `arkmdxp4.dll` thread as "padding" the game expects (I actually tried reducing the ring buffers to *only* write entries on state transitions, but it broke gameplay, which means the game still depends regular reporting of the pad state). Here's an example of how a ring buffer would look if an arrow was pressed at some time 32ms (assume 165Hz game refresh rate, ~6ms updates):

```
Old (only frame updates):
1 | 36ms
0 | 30ms
0 | 24ms
0 | 18ms
0 | 12ms
0 | 6ms
0 | 0ms
```
The game only sees the press event at time 36ms on frame update, so time 36ms is assigned to the press time of that arrow (4ms later than the real press time).

```
New (frame updates + event updates from rawinput.cpp):
1 | 36ms
1 | 32ms <- from rawinput.cpp
0 | 30ms
0 | 24ms
0 | 18ms
0 | 12ms
0 | 6ms
```
The game sees the real press time 32ms on frame update, so the correct press time 32ms is assigned.

## Testing
- Played on various refresh rates and verified the controller input remains highly responsive.
- Inspected the ring buffers with a debugger and verified entries for press and release events from `rawinput.cpp` were correctly being added for their respective players.

### DIFF
--- a/src/spice2x/acio/mdxf/mdxf.cpp
+++ b/src/spice2x/acio/mdxf/mdxf.cpp
@@ -7,7 +7,6 @@
 #include "rawinput/rawinput.h"
 #include "util/logging.h"
 #include "util/utils.h"
-#include "launcher/logger.h"
 #include <mutex>
 
 // constants


.gitignore
```diff
@@ -10,4 +10,4 @@ cmake-build*
 
 # user config
 build_all.local.sh
-build_docker.local.sh
+build_docker.local.sh
```
bicarus-dev: please revert

src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -189,29 +212,64 @@ static bool __cdecl ac_io_mdxf_update_control_status_buffer(int node) {
...
     return true;
 }
 
+static bool __cdecl ac_io_mdxf_update_control_status_buffer(int node) {
+	return ac_io_mdxf_update_control_status_buffer_impl(node, ARKMDXP4_POLL);
+}
+
+// Used for triggering updates of the controller states from outside arkmdxp4.dll main refresh loop (i.e. within rawinput.cpp on controller events)
+void mdxf_poll() {
+	ac_io_mdxf_update_control_status_buffer_impl(17, EXTERNAL_POLL);
```
bicarus-dev: This routine is called for every game, on every poll. Instead, it should only do work if this I/O module has in active use by MDX.

src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -126,38 +147,40 @@ static bool __cdecl ac_io_mdxf_set_output_level(unsigned int a1, unsigned int a2
...
-            return false;
-    }
-    memset(buffer, 0, STATUS_BUFFER_SIZE);
+	uint8_t (*buffer)[STATUS_BUFFER_SIZE];
+	uint8_t *head = nullptr;
+	uint16_t *prev_state = nullptr;
+	std::mutex* mutex = nullptr;
+	
+	switch (node) {
+		case 17:
```
bicarus-dev: I didn't notice until now but all of these should be 4 spaces, not tabs 🙂

src/spice2x/acio/mdxf/mdxf.cpp
cchike: There's an `if (avs::game::is_model("MDX"))` check within `ac_io_mdxf_update_control_status_buffer_impl`. Should I add one here too?

src/spice2x/acio/mdxf/mdxf.cpp
cchike: Are you saying indentation should be done with spaces and not tabs throughout the whole file?

src/spice2x/acio/mdxf/mdxf.cpp
bicarus-dev: That's better, but it would still mean that this poll function is doing work for all MDX versions, regardless of what I/O module is in use. (The fact that all I/O modules are active by default is an odd design quirk of spice..)

Maybe there should be a global that starts out as FALSE but flips to TRUE when at least one call is made to `ac_io_mdxf_update_control_status_buffer_impl`, which implies the game is using it.

src/spice2x/acio/mdxf/mdxf.cpp
bicarus-dev: Yes, 4 spaces is the standard in this repo, no tabs.

.gitignore
cchike: Done

src/spice2x/acio/mdxf/mdxf.cpp

src/spice2x/acio/mdxf/mdxf.cpp

PR #450, opened 2025-12-15
MDX: Added Backfill Logic For Ring Buffer Health (P4IO Timing Fix) 

cchike:
## Context
While testing the emulator on 60Hz for a while, I noticed there was a pretty consistent timing bug with steps that had quick releases in between them (mainly footswitches and sometimes jacks). As mentioned in my previous commits, `arkmdxp4` determines press/release times for steps based on a 7-poll history on every frame, which is normally populated at ~250Hz from the MDXF board in `libacio2`, yielding around 28ms of polling history. When emulated, this is instead being populated internally from `arkmdxp4` once per frame, meaning the 7-poll history could have up to `1000/60 * 7 = 117ms` worth of step history at 60Hz. In the case of a footswitch/jack, if you account for the real-time press and release event that's inserted from `rawinput` during that time, there'd actually be up `1000/60 * (7-2) = 83ms` worth of step history. This is still a large enough window of time for `arkmdxp4` to start confusing older on-state polls with more recent ones. Here is a real ring buffer I logged when I reproduced the bug:

```
Frame #6: 82ms | 1
          70ms | 1 <- rawinput real-time press
Frame #5: 66ms | 0
Frame #4: 49ms | 0
Frame #3: 34ms | 0 <- rawinput real-time release (coincided with frame update)
Frame #2: 16ms | 1 <- stale "held" poll
Frame #1:  0ms | 1 <- stale "held" poll
```

When `arkmdxp4` determines that a new press event occurred, it uses the timestamp of the oldest on-state poll in the ring buffer as the press time. In this case, it uses time 0ms instead of the correct 70ms, resulting in the game thinking the player stepped 70ms before they actually did. For reference, that same ring buffer snapshot would've looked like this in the real implementation:

```
Frame #6: 82ms | 1
          78ms | 1
          74ms | 1
          70ms | 1
Frame #5: 66ms | 0
          62ms | 0
          58ms | 0
```

The correct time 70ms would be assigned to that step in this case. This demonstrates the importance of keeping the ring buffers filled with only recent poll history, closer to the threshold of around ~28ms that the real implementation expects.

## Description of change
~~When the module is booted, it counts how many times `ac_io_mdxf_update_control_status_buffer` was called in the first three seconds, then calculates the approximate refresh rate of the current game instance. If the refresh rate is under `THRESHOLD_REFRESH_RATE` (120Hz), a low-priority thread dedicated to inserting polls into the ring buffers is started and run at `THREAD_REFRESH_RATE_HZ` (125Hz). Combined with the polls inserted by the main `arkmdxp4` thread and `rawinput` real-time events, the 125Hz update cadence is enough to bring the span of poll history represented by the ring buffers closer to parity with what `arkmdxp4` expects.~~

EDIT: The prior thread implementation was replaced with new logic that, on each call to the update function, backfills the ring buffer with entries representing the last known state of the controller. For example, given the ring buffer with a single entry:

```
0ms | 0
```
When the update function is called on the next frame at 60Hz, the function will insert intermediate entries as "padding" before updating to its current state:

```
16ms | 1 <- state at current frame
12ms | 0 <- "padding"
8ms  | 0 <- "padding"
4ms  | 0 <- "padding"
0ms  | 0 <- last known state
```

## Considerations
~~In the future, it might be worth exploring a route that interpolates polls instead of relying on a thread for inserting polls (in other words, creating intermediate poll entries based on the known real-time press and release events from `rawinput`). For example, if `rawinput` reported a press at time 10ms and a release at time 30ms, it's known that any poll in between those times would report an on-state for that arrow. However, considering this would involve adding multiple entries per call (for times in the past) and how the update function can be called from two separate threads, it may be introduce hard-to-debug edge cases and add unnecessary complexity overall.~~

EDIT: Implemented

## Testing
~~-Verified the thread was inserting polls at the expected cadence
-Verified the thread is only created when the game's refresh rate is below the threshold
-Verified the thread introduced no significant increase in CPU~~
- Verified the entries were being backfilled correctly on several different refresh rate instances
- Verified the bug could no longer be reproduced

### DIFF
--- a/src/spice2x/acio/mdxf/mdxf.cpp
+++ b/src/spice2x/acio/mdxf/mdxf.cpp
@@ -259,6 +259,7 @@ static bool __cdecl ac_io_mdxf_update_control_status_buffer_impl(int node, MDXFP
         }
         
         // The start and stop time cutoffs for backfilling entries. Min(..) ensures times aren't negative.
+        // The stop time is just before the current_time, set by BACKFILL_PADDING_MS, which avoids the last backfilled entry being too close to current_time.
         uint64_t start_time = *prev_time;
         const uint64_t stop_time = current_time - std::min<uint64_t>(current_time, BACKFILL_PADDING_MS);
         
@@ -288,7 +289,7 @@ static bool __cdecl ac_io_mdxf_update_control_status_buffer_impl(int node, MDXFP
             
             time += BACKFILL_INTERVAL_MS;
             
-            // If the current time is reached, then write current_time and current_state instead for this final iteration
+            // If the stop time is reached, then write current_time and current_state instead for this final iteration
             const bool isEdge = (time >= stop_time);
             if (isEdge) {
                 state = current_state;


src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -56,6 +72,86 @@ static uint64_t arkGetTickTime64() {
...
+    if (!MDXF_THREAD_RUNNING.compare_exchange_strong(expected, true)) {
+        return;
+    }
+
+    MDXF_THREAD = std::thread([] {
+        SetThreadPriority(GetCurrentThread(), THREAD_PRIORITY_BELOW_NORMAL);
+
+        while (MDXF_THREAD_RUNNING.load(std::memory_order_acquire)) {
+            mdxf_poll(false);
+            std::this_thread::sleep_for(THREAD_PERIOD);
```
bicarus-dev: This is not going to be reliable; sleeping a thread guarantees that it will sleep at least THREAD_PERIOD but can be longer, since it relies on the OS's Sleep function, the minimum sleep duration is platform-specific, and so on. In practice this means any sleep less than 10ms is going to sleep for around 10ms, but can be as high at 15.6ms. In general you want to avoid writing a loop that relies on the scheduler running your thread at precise intervals.

cchike: In that case, I think I might be better off going with my alternative non-thread-based solution I mentioned in my comment above. Should I make the changes here or make a new PR?

src/spice2x/acio/mdxf/mdxf_poll.h
```diff
@@ -2,4 +2,4 @@
 #pragma once
 
 // Called from rawinput thread whenever inputs have just been updated
-void mdxf_poll();
\ No newline at end of file
+void mdxf_poll(bool isExternal);
```
cchike: @bicarus-dev, I was meaning to ask: regarding the place I put this  `mdxf_poll()` in `rawinput.cpp`, is this guaranteed to be the most "real-time" location to log when the controller's input happens? Is this in a high priority thread? I just want to make sure the least amount of precision is lost from when the controller's input hits the OS and when the timestamp for it is generated. For example, if I'm using a 1000Hz controller, I want that precision to propagate to my `mdxf_poll()` function.

I ask this because in my testing on 60Hz, it still feels like precision is lost somewhere along the way even though the state change and input time is inserted into the ring buffers immediately

bicarus-dev: yes, the rawinput thread runs at realtime priority and it's as fast it's going to get

src/spice2x/acio/mdxf/mdxf.cpp
bicarus-dev: I'm trying to understand the issue... 

if the problem is that the game is seeing "stale" entries because it looks back at N past timestamps, couldn't each rawinput update insert more than one event? In other words, couldn't "step on" event from rawinput thread cause "off, off, off, on" events to be inserted so that you effectively catch up to the polls that were missed in the last frame?

bicarus-dev: In other words,

1. The raw input thread ALWAYS has the most up-to-date state on whether an arrow is stepped ON or OFF, since it's what Windows tells us.

1. Your new periodic worker in this thread is just reading the cached value from Buttons API; i.e., you're just reading in the old value and inserting them into the queue.

You could forgo 2 and just have 1 insert more entries in to the ring buffer. Since the entries are timestamped, you don't have to insert it periodically, you can just lie and make up a the timestamp so that it's in the past, but not before the last poll from the game.

cchike: I agree that would be better and that’s essentially what my alternative solution is doing actually

cchike: Should I update this PR with the changes?

bicarus-dev: Up to you, you can keep this or open a new PR, fine either way.

src/spice2x/acio/mdxf/mdxf_poll.h
cchike: On a related note, do you know if it’s feasible to internally run the game at a refresh rate higher than what the monitor is capable of? For example, running the game at 120Hz but with the 60Hz monitor only showing half the frames? I feel like the game running at a lower refresh rate has a tangible effect on timing precision beyond visual latency.

bicarus-dev: I don't think so, that'll almost certainly require hacking the game

src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -19,11 +19,15 @@ static uint8_t HEAD_P2 = 0;
...
+static uint64_t PREV_TIME_P1 = 0;
+static uint64_t PREV_TIME_P2 = 0;
 
 static std::mutex MUTEX_P1;
 static std::mutex MUTEX_P2;
 
-static bool MDXF_IS_ACTIVE = false;
+static bool IS_MDXF_ACTIVE = false;
+
+static uint8_t BACKFILL_INTERVAL_MS = 4;
```
bicarus-dev: `const`

src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -236,29 +243,56 @@ static bool __cdecl ac_io_mdxf_update_control_status_buffer_impl(int node, MDXFP
...
         
-        //Write current game time
-        *(uint64_t*)&buffer_entry[0x18] = arkGetTickTime64();
+        // Ensures the first iteration will write the first entry at current_time and not backfill to time 0ms. Min(..) ensures time isn't negative.
+        if (time == 0) {
+            time = current_time - std::min<uint64_t>(current_time, BACKFILL_INTERVAL_MS);
+        }
+        
+        // Backfill entries a fixed interval apart from each other between prev_time and current_time
+        while (time < current_time) {
```
bicarus-dev: do you need some kind of upper bound on how many entries we fill out as a precaution? maybe cap to writing STATUS_BUFFER_NUM_ENTRIES at most?

e.g., if current_time is significantly larger than time, then we might loop here over and over again overwriting to the buffer

src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -236,29 +243,56 @@ static bool __cdecl ac_io_mdxf_update_control_status_buffer_impl(int node, MDXFP
         }
         
         uint16_t current_state = (uint16_t(up_down) << 8) | left_right;
+        uint64_t current_time = arkGetTickTime64();
```
bicarus-dev: `const`

src/spice2x/acio/mdxf/mdxf.cpp
cchike: `time` is instantiated as `*prev_time`, so the difference between that and `current_time` is always exactly equal to the elapsed time since the update function was last run, i.e. the duration of a frame at most.

src/spice2x/acio/mdxf/mdxf.cpp

src/spice2x/acio/mdxf/mdxf.cpp

src/spice2x/acio/mdxf/mdxf.cpp
bicarus-dev: right, but we should assume the worst case scenario and not assume that the game will call you every frame, since this is running as part of a thread that processes window messages, and blocking that is catastrophic 

(It depends on game, and I haven't verified it for DDR, but some games will straight up not bother with i/o if you're in test menu, if the game is between loading screens, etc)

cchike: FWIW, the IO is constantly being read on every frame. Any breakpoint I've ever set in the normal IO pipeline triggers immediately, while on any screen. I can add something though, it's not a problem

bicarus-dev: Thanks!

src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -235,30 +242,60 @@ static bool __cdecl ac_io_mdxf_update_control_status_buffer_impl(int node, MDXFP
...
+        uint64_t time = *prev_time;
         
-        //Write current game time
-        *(uint64_t*)&buffer_entry[0x18] = arkGetTickTime64();
+        // Ensures the first iteration will write the first entry at current_time and not backfill to time 0ms. Min(..) ensures time isn't negative.
+        if (time == 0) {
+            time = current_time - std::min<uint64_t>(current_time, BACKFILL_INTERVAL_MS);
+        }
+        
+        uint16_t entries_written = 0;
```
bicarus-dev: much simpler way to do this would be to ensure that `time` is not too far from `current_time`; i.e., if `current_time - time` is too large, you cap it to `time = current_time - some_fixed_multiple_of_BACKFILL_INTERVAL_MS`

bicarus-dev: Thanks!

src/spice2x/acio/mdxf/mdxf_poll.h
cchike: @bicarus-dev, through `get_performance_milliseconds()` logging I've observed that there's a fractional amount of varying latency between when `mdxf_poll()` is called and when `mdxf_poll()` actually runs, which is removing at least 1ms precision from the input time. I verified an improvement in timing precision once I created the `arkGetTickTime64()` timestamp here and passed it along. Do you have a suggestion of how I can cleanly expose that function here to get the precise time the key event happened and pass that into `mdxf_poll()`, instead of generating that timestamp inside the update function?

bicarus-dev: I think we should capture a timestamp in the beginning of `input_wnd_proc` for `WM_INPUT` case and propagate it. That's basically the earliest we get a notification from Windows that a raw input event happened.

cchike: Right, but the timestamp needs to be in `arkmdxp4`’s timeframe from `arkGetTickTime64()`. I was asking if I can import it in a way that follows protocol.

bicarus-dev: Oh, right, I see what you're saying. There isn't an established pattern because this the only I/O module that we implement that is timestamped.  You can just add a new function like `mdxf_timestamp()` and call it there, and stash the value in a global, or propagate the value on the stack.

bicarus-dev: While we're here, one other thing I noticed - `mdxf_poll` is being called inside the for loop that iterates through devices. I don't think that's necessary, you probably only need to do it once per message.

cchike: > Oh, right, I see what you're saying. There isn't an established pattern because this the only I/O module that we implement that is timestamped. You can just add a new function like `mdxf_timestamp()` and call it there, and stash the value in a global, or propagate the value on the stack.

Just to be clear, are you suggesting something can be called from within `input_wnd_proc`? Because I’m hoping that can be done directly with `arkGetTickTime64()`.

> While we're here, one other thing I noticed - `mdxf_poll` is being called inside the for loop that iterates through devices. I don't think that's necessary, you probably only need to do it once per message.

Are you sure? From my timestamp logging at that call location, the logs were only triggered once per event.

bicarus-dev: I'm suggesting that early in `input_wnd_proc` you call `mdxf_timestamp` which calls `arkGetTickTime64`. Would that resolve the issue?

For the second problem, it would only show up if you have multiple rawinput devices bound.

cchike: > I'm suggesting that early in `input_wnd_proc` you call `mdxf_timestamp` which calls `arkGetTickTime64`. Would that resolve the issue?

I don’t think so because the initial problem is that the latency comes immediately upon calling `mdxf_poll` before any work is even done. I don’t know why that is, but that’s reason I was hoping to just circumvent putting the timestamp call in a wrapper and get it straight from the source

bicarus-dev: Are you saying that from beginning of `mdxf_poll` to end of `mdxf_poll` is taking a millisecond? The only thing I can think of is that there is lock contention on the mutex you are using, or that `GameAPI::Buttons::getState` is taking too long.

Does it really matter though - as long as you have accurate timestamp captured early in the window message, it should be accurately accounted for in the game engine, right?

bicarus-dev: I'm doing some performance logging as well, but my feeling is that you have the ARKMDXP4_POLL and EXTERNAL_POLL occasionally fighting to acquire the mutex at the same time, one of them loses, so it gets suspended for a bit, so you notice a delay.

`GameAPI::Buttons::getState` also has its own mutex (per device) so that is also a potential source of contention.

cchike: Regarding the latency I was observing, I must've logged something wrong because now I'm seeing that the call to `mdxf_poll()` is pretty much instant, as I would've expected. So I'm just generating the timestamp for `arkGetTickTime64()` at the very beginning of `mdxf_poll()` and passing it along.

 > `GameAPI::Buttons::getState` also has its own mutex (per device) so that is also a potential source of contention.

Is there any state value I could potentially directly pass into `mdxf_poll()` from `rawinput` to prevent the need to read the state object again?

bicarus-dev: You could call `Button.GetLastState` which would just give you the previous value without checking the devices.

Right now I'm experimenting with a try-acquire lock for the mutex to avoid the contention altogether.

cchike: I'm also thinking I might be able to get away with not reading the state object at all on the `ARKMDXP4_POLL` update cycles and just populate the ring buffer entry for those calls using the previous known state instead, since every nontrivial controller state change will be coming from `rawinput` anyways

bicarus-dev: Unfortunately there are some edge cases where input can be provided via spice API for example, in that case it wouldn't come from rawinput.

I think it would be better if the ARMDXP4_POLL just gave up and didn't update the ring buffer if we know that it was updated recently by rawinput (in the last couple ms), or if rawinput is literally in the middle of doing so (e.g., via unique_lock try_acquire)

cchike: Hold on, are you saying there are controller inputs that can randomly not get processed where I'm calling `mdxf_poll()` in `rawinput`?

bicarus-dev: No, I mean inputs *not* from a controller, like a third party app that interacts over the SpiceAPI.

cchike: > No, I mean inputs _not_ from a controller, like a third party app that interacts over the SpiceAPI.

I'm not following how this would relate to pad input during gameplay

cchike: > You could call `Button.GetLastState` which would just give you the previous value without checking the devices.

Would this functionally be the same for our use case here? How exactly do the two calls differ?

bicarus-dev: > I'm not following how this would relate to pad input during gameplay

For example, spicecompanion on Android has a controller mode that shows a DDR controller on a phone screen and people can send pad input that way. Obscure? Yes. Should it continue to be supported? Yes

> > You could call `Button.GetLastState` which would just give you the previous value without checking the devices.
> Would this functionally be the same for our use case here? How exactly do the two calls differ?

GetState checks what device is bound to the button, checks the last-reported rawinput state for that device, and calculates the on/off state

GetLastState just returns you the most recent result from the last call to GetState

bicarus-dev: I think the lock contention issue would be avoided if you removed the lock_guard and instead did this early in `ac_io_mdxf_update_control_status_buffer_impl`

```c++
        std::unique_lock<std::mutex> lock(*mutex, std::try_to_lock);
        if (source == ARKMDXP4_POLL && !lock.owns_lock()) {
            return true;
        }
```

this way, the rawinput thread gets priority, and if the game happens to poll for input, you do nothing and return. I think this would work, although you might have a better idea since you looked into the game's I/O logic more closely. (I haven't actually tried this yet)

cchike: > I think the lock contention issue would be avoid if you removed the lock_guard and instead did this early in `ac_io_mdxf_update_control_status_buffer_impl`
> 
> ```c++
>         std::unique_lock<std::mutex> lock(*mutex, std::try_to_lock);
>         if (source == EXTERNAL_POLL && !lock.owns_lock()) {
>             return true;
>         }
> ```

I'm not quite sure if it's the locks within the update methods that are the problem. I actually logged how long it took to execute the `getState` block after you mentioned it and it was regularly 0.3-0.7ms alone, compared to the actual ring buffer writing which was on the order of <1µs. I think if there was a way around brute force reading of the controller state it would solve part of the latency issue.

Also, regarding that block of code, did you mean to put `ARKMDXP4_POLL` instead? We need the external polls to execute because that's where the real-time information comes from

bicarus-dev: But what is the latency issue though? If you acquire the timestamp before checking the buttons, and put that timestamp into the ring buffer, doesn't the game understand that the input happened in the past, so the getState taking 0.7ms doesn't really matter?

cchike: > But what is the latency issue though? If you acquire the timestamp before checking the buttons, and put that timestamp into the ring buffer, doesn't the game understand that the input happened in the past, so the getState taking 0.7ms doesn't really matter?

It’s worth considering that the update function is called twice on the main thread, once for each player, so it’s actually blocking for twice that amount of time. In the real non-emulated implementation, the update function just marks the head of the ring buffer as consumed and doesn’t do any actual work. I can’t help but think the game’s internal timing mechanism might start acting up if the main thread loop takes an additional 1-2ms to complete, even if the timestamps are accurate in theory.

bicarus-dev: To add one more clarification - `Buttons::getState`  taking 0.7ms would also be a lock contention issue, because each device has a lock that is acquired by the writer (the rawinput thread) and the two readers (`ac_io_mdxf_update_control_status_buffer_impl`).  If you make `ARKMDXP4_POLL` thread give up, then you have no contention whatsoever, because the rawinput thread is both the reader and the writer.

My suggestion is to move the lock so that it covers all of the operation in `ac_io_mdxf_update_control_status_buffer_impl` (both the getState checks and also the updating of the buffers). 

>Also, regarding that block of code, did you mean to put ARKMDXP4_POLL instead? We need the external polls to execute because that's where the real-time information comes from

Sorry, you're right, I made the correction.

bicarus-dev: > It’s worth considering that the update function is called twice on the main thread, once for each player, so it’s actually blocking for twice that amount of time. In the real non-emulated implementation, the update function just marks the head of the ring buffer as consumed and doesn’t do any actual work. I can’t help but think the game’s internal timing mechanism might start acting up if the main thread loop takes an additional 1-2ms to complete, even if the timestamps are accurate in theory.

Seems like we posted at the same time. Please disregard my reply above.

I think what you should try is.

* forget this spiceapi stuff for now, assume for now that rawinput is the only source
* in the rawinput thread (EXTERNAL_POLL case), call GetState and fill out the buffer
* in the game-driven poll (ARKMDXP4_POLL case), do the minimum required to maintain the buffer, and if you must, call GetLastState to add a new entry in the ring buffer

That would just work, right?  And we can worry about how to solve the non-rawinput case separately (might be an easier problem as we can disregard the latency issue since you're probably not playing seriously anyway if you're using the API)

I really don't think the `GetState` calls would take more than 0.1ms if you have zero lock contention. I believe that the 0.3ms-0.7ms you are seeing is due to the two threads racing to read/write states at the same time.

cchike: > Is there any state value I could potentially directly pass into `mdxf_poll()` from `rawinput` to prevent the need to read the state object again?

Just wanted to ask this again to make sure this couldn't be an option. It seems wasteful to process the input then have to make a `getState` call separately to get that same input.

> I really don't think the GetState calls would take more than 0.1ms if you have zero lock contention. I believe that the 0.3ms-0.7ms you are seeing is due to the two threads racing to read/write states at the same time.

Even when I switched the implementation to only call `getState` on the `rawinput` event polls, it's still blocking for some amount of variable time between 0-1ms

bicarus-dev: > > Is there any state value I could potentially directly pass into `mdxf_poll()` from `rawinput` to prevent the need to read the state object again?
> 
> Just wanted to ask this again to make sure this couldn't be an option. It seems wasteful to process the input then have to make a `getState` call separately to get that same input.

I think I understand what you're trying to say (can I reach around and inspect the HID state of devices while I'm updating them) but it doesn't quite work because you don't know which button is bound to which input (you don't know which device and which value you need to look for 1P_DOWN), without breaking through multiple layers of abstraction. So the short answer is no, not really, there isn't a clean shortcut you can take, I would stick to calling `Buttons::GetState` even if it seems expensive.

I know the rawinput stuff is severely over-engineered but that's the code I inherited... All this complexity exists because you can have N:N mapping between button and input (a button on a controller can be bound to multiple game inputs, and multiple buttons can be bound to one game input), plus all the API stuff.

cchike: > I think I understand what you're trying to say (can I reach around and inspect the HID state of devices while I'm updating them) but it doesn't quite work because you don't know which button is bound to which input (you don't know which device and which value you need to look for 1P_DOWN), without breaking through multiple layers of abstraction. So the short answer is no, not really, there isn't a clean shortcut you can take, I would stick to calling `Buttons::GetState` even if it seems expensive.

At the very least, is there a way to differentiate between 1P and 2P input? Right now, any `rawinput` event calls the update method for both 1P and 2P to check if the controller state changed

cchike: Side question: do naive vs non-naive bindings have any effect on latency in general?

bicarus-dev: > At the very least, is there a way to differentiate between 1P and 2P input? Right now, any `rawinput` event calls the update method for both 1P and 2P to check if the controller state changed

No way to make a distinction like this because we don't assume that one HID device is mapped to 1p and another is mapped to 2p.

For naive vs. non-naive, in spice there is absolutely less code and less locking we do, but in Windows rawinput is *supposed* to be faster. So without doing performance measurements I can't say what's faster.

cchike: > For naive vs. non-naive, in spice there is absolutely less code and less locking we do, but in Windows rawinput is _supposed_ to be faster. So without doing performance measurements I can't say what's faster.

I'm not super familiar with the difference besides non-naive being tied to the specific device and naive being device-agnostic. Does only one or the other use `rawinput`? If so, what does the other use?

bicarus-dev: Naive = calls `GetAsyncKeyState` to get keyboard state (and we have no clue how quickly Windows updates this)
Bind = win32 rawinput API

bicarus-dev: By the way, I instrumented the code like this:

```c++
ac_io_mdxf_update_control_status_buffer_impl()
{
...
        const auto start = get_performance_milliseconds();

        if (GameAPI::Buttons::getState(RI_MGR, buttons.at(button_map[0]))) {
            up_down |= 0xF0;
        }
        if (GameAPI::Buttons::getState(RI_MGR, buttons.at(button_map[1]))) {
            up_down |= 0x0F;
        }
        if (GameAPI::Buttons::getState(RI_MGR, buttons.at(button_map[2]))) {
            left_right |= 0xF0;
        }
        if (GameAPI::Buttons::getState(RI_MGR, buttons.at(button_map[3]))) {
            left_right |= 0x0F;
        }

        log_misc("ddrtime", "{}", get_performance_milliseconds() - start);
```

and with two xbox controllers I do not ever see this number go above 0.002 (2 microseconds)... how did you end up getting sub-millisecond? If you are seeing that, it's a bit concerning because it means we are barely keeping up with 1000Hz HID devices.

bicarus-dev: Let's use this Issue instead for discussion - https://github.com/spice2x/spice2x.github.io/issues/452

Issue #452, opened 2025-12-16
discussion: DDR p4io timing issues


bicarus-dev:
Continuation of discussion from https://github.com/spice2x/spice2x.github.io/pull/450#discussion_r2622970956

Outstanding topics:

- [ ]  cchike: should move timestamp to be earlier in the call for more accurate timestamping

- [ ]  cchike: seeing long runtime of Buttons API https://github.com/spice2x/spice2x.github.io/pull/450#discussion_r2622477080
>I actually logged how long it took to execute the getState block after you mentioned it and it was regularly 0.3-0.7ms alone, compared to the actual ring buffer writing which was on the order of <1µs. I think if there was a way around brute force reading of the controller state it would solve part of the latency issue. 

but bicarus not seeing the problem https://github.com/spice2x/spice2x.github.io/pull/450#discussion_r2622970956
>...and with two xbox controllers I do not ever see this number go above 0.002 (2 microseconds)... how did you end up getting sub-millisecond? If you are seeing that, it's a bit concerning because it means we are barely keeping up with 1000Hz HID devices.

(FWIW, I later tried this with my razer keyboard which polls at 1000hz in HID mode, seems to make no meaningful difference...)

- [ ] bicarus: can we move the call to mdxf_poll out of the loop so that it's not called multiple times (in the case where there are multiple rawinput devices in use)

- [ ] bicarus: ARKMDXP4_POLL thread and EXTERNAL_POLL threads may run into lock contention, should ARKMDXP4_POLL  thread try-acquire the lock and give up if it's already in use? alternatively, should ARKMDXP4_POLL  thread never use GetState directly but always use GetLastState to avoid checking HID device state?

bicarus-dev:
How did you measure the latency? I can't imagine that a call to `GetAsyncKeyState` would take long. It's less than a microsecond on my PC.

bicarus-dev:
Is your cab on Win7 by any chance? I wonder if GetAsyncKeyState is way slower on older OS. I was testing on Win10 and 11.

If you're not getting great results, maybe you can revive the 125Hz worker thread but use the following to ensure that you get finer granularity for timers:

* https://learn.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timebeginperiod
* see the Remarks section in particular about Win10 and Win11 specifics

as the documentation says timeBeginPeriod has effect on the scheduler so it *may* affect other parts of DDR and could in theory affect timing. It would be super subtle though.

[...]

PR #456, opened 2025-12-17
MDX: Updated Ring Buffer Backfill Implementation (P4IO Timing Fix)

cchike:
#452  

## Description of change
- Extension of #450 
- Moved the call to `arkGetTickTime64()` outside the update function to bring the timestamps closer to real-time
- Changed the update function to only call `getState()` if called from `rawinput`, otherwise just fill the ring buffer entries with the last known state
- Changed the update function to only write one ring buffer entry if called from `rawinput`
- Moved the call to `mdxf_poll()` within `rawinput.cpp` outside the loop that iterates over all devices

## Testing
- Logging was done to verify the span of poll history covered by each request for 7 ring buffer entries was ~28ms, as the game expects
- Observed some improvement in timing precision since the prior implementation

### DIFF
--- a/src/spice2x/acio/mdxf/mdxf.cpp
+++ b/src/spice2x/acio/mdxf/mdxf.cpp
@@ -245,8 +245,7 @@ static bool __cdecl ac_io_mdxf_update_control_status_buffer_impl(int node, MDXFP
                 left_right |= 0x0F;
             }
             current_state = (uint16_t(up_down) << 8) | left_right;
-        }
-        else {
+        } else {
             current_state = *prev_state;
         }
         


src/spice2x/rawinput/rawinput.cpp
```diff
@@ -1826,10 +1826,10 @@ LRESULT CALLBACK rawinput::RawInputManager::input_wnd_proc(
 
                 // free device
                 device.mutex->unlock();
-                
-                // update controller state ring buffers (DDR/MDXF)
-                mdxf_poll();
             }
+            
+            // update controller state ring buffers (DDR/MDXF)
+            mdxf_poll();
```
bicarus-dev, line 1832: turns out my previous concern was moot since the loop above is only supposed to run once anyway (there really should be a `break;` at the end, I guess the original devs missed it) but thanks for making this change regardless, it's better this way even if it has no functional impact after all..

src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -224,37 +224,39 @@ static bool __cdecl ac_io_mdxf_update_control_status_buffer_impl(int node, MDXFP
...
+                left_right |= 0xF0;
+            }
+            if (GameAPI::Buttons::getState(RI_MGR, buttons.at(button_map[3]))) {
+                left_right |= 0x0F;
+            }
+            current_state = (uint16_t(up_down) << 8) | left_right;
         }
-        if (GameAPI::Buttons::getState(RI_MGR, buttons.at(button_map[3]))) {
-            left_right |= 0x0F;
+        else {
```
bicarus-dev: nit: `} else {`

PR #460, opened 2025-12-18
MDX: Merged Thread and Backfill Implementations (P4IO Timing Fix)

cchike:
#452 

## Description of change
(Another extension of #450)
Merged the old threading implementation into the current backfill implementation because it was showing signs of having slightly better timing performance. The following description is taken from the strikethrough text from the referenced PR above:
>When the module is booted, it counts how many times ac_io_mdxf_update_control_status_buffer was called in the first three seconds, then calculates the approximate refresh rate of the current game instance. If the refresh rate is under THRESHOLD_REFRESH_RATE (120Hz), a low-priority thread dedicated to inserting polls into the ring buffers is started and run at THREAD_REFRESH_RATE_HZ (125Hz). Combined with the polls inserted by the main arkmdxp4 thread and rawinput real-time events, the 125Hz update cadence is enough to bring the span of poll history represented by the ring buffers closer to parity with what arkmdxp4 expects.

The default mode is `THREAD_MODE` but can be switched to `BACKFILL_MODE` (which could possibly be made as a `spicecfg` option).

There was also a fixed bug in the `ac_io_mdxf_get_control_status_buffer` function, which should not have been writing to the `head` pointer. This was allowing for an edge case situation where the `head` pointer was advanced in another thread then rewound back to the previous `head` in this thread. The real implementation only reads the pointer and returns it.

## Testing
- Logged the ring buffer states after many iterations and verified regular structure
- Same playtesting as the prior PRs

### DIFF
--- a/src/spice2x/acio/mdxf/mdxf.cpp
+++ b/src/spice2x/acio/mdxf/mdxf.cpp
@@ -33,7 +33,7 @@ static const uint8_t BACKFILL_PADDING_MS = 2;
 // These are used to determine if a thread needs to be spun up to keep pad state ring buffers populated with enough recent polls
 static uint64_t START_TIME = 0;
 static int CALL_COUNT = 0;
-static int THRESHOLD_REFRESH_RATE = 120;
+static const int THRESHOLD_REFRESH_RATE = 120;
 static bool IS_REFRESH_RATE_DETERMINED = false;
 static bool IS_THREAD_NEEDED = false;
 


src/spice2x/acio/mdxf/mdxf.cpp
```diff
@@ -61,6 +87,86 @@ static uint64_t arkGetTickTime64() {
...
 
+// Used to keep the ring buffer populated with steady updates. 60Hz interval is too slow
+static void mdxf_thread_start() {
+    bool expected = false;
+    if (!MDXF_THREAD_RUNNING.compare_exchange_strong(expected, true)) {
+        return;
+    }
+
+    MDXF_THREAD = std::thread([] {
+        SetThreadPriority(GetCurrentThread(), THREAD_PRIORITY_BELOW_NORMAL);
```
bicarus-dev, line 98: the priority could be higher, not sure if that's better or worse, but there isn't a great reason to keep this at below-normal

bicarus-dev: if it works fine at below normal that's fine too of course

cchike: The state or exact time values that come from the ring buffer entries generated by that thread don't really matter so I thought there'd be nothing much to lose by ensuring it uses the least amount of resources necessary

PR #467, opened 2025-12-22
MDX: Refactored p4io.cpp

cchike:
#452 

## Description of change
Added `Sleep(4)` call to regulate how fast the `device_io` function will call `getState` for each button. The panel buttons were commented out because those are handled by the MDXF Control Status Buffer functions anyways, so the work was redundant. It's also why the timing precision from this function doesn't matter, so it can afford a longer sleep

## Testing
- Verified IO still functions properly
- Verified decrease in overall CPU usage

### DIFF
--- a/src/spice2x/games/ddr/p4io/p4io.cpp
+++ b/src/spice2x/games/ddr/p4io/p4io.cpp
@@ -254,7 +254,7 @@ namespace games::ddr {
             switch((dwIoControlCode>>2)&0x3FFF) {
                 case P4IO_IOCTL_GET_INPUTS: {
                     
-                    // cool down
+                    // Prevents this function from being called at its normal 2000-2500 kHz cadence and overloading the CPU, instead reduces it to 250 Hz
                     Sleep(4);
                     
                     memset(lpOutBuffer, 0, 16);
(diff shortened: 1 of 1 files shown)


src/spice2x/games/ddr/p4io/p4io.cpp
```diff
@@ -253,28 +253,33 @@ namespace games::ddr {
         if(((dwIoControlCode>>16)&0xFFFF) == FILE_DEVICE_UNKNOWN) {
             switch((dwIoControlCode>>2)&0x3FFF) {
                 case P4IO_IOCTL_GET_INPUTS: {
+                    
+                    // cool down
+                    Sleep(4);
```
bicarus-dev, line 258: how often is this called normally?

src/spice2x/games/ddr/p4io/p4io.cpp
```diff
@@ -253,28 +253,33 @@ namespace games::ddr {
         if(((dwIoControlCode>>16)&0xFFFF) == FILE_DEVICE_UNKNOWN) {
             switch((dwIoControlCode>>2)&0x3FFF) {
                 case P4IO_IOCTL_GET_INPUTS: {
+                    
+                    // cool down
```
bicarus-dev: could use some comment here to explain why we need to slow down (i.e., the game calls this too often)

src/spice2x/games/ddr/p4io/p4io.cpp
```diff
@@ -253,28 +253,33 @@ namespace games::ddr {
...
 
                     const std::pair<size_t, uint8_t> buttonMapping[] =
                     {
+                        // Panel input comes from MDXF Control Status Buffer functions
                         {Buttons::SERVICE, 25}, //Good
                         {Buttons::TEST, 28}, //Good
                         {Buttons::COIN_MECH, 24}, //Good
                         {Buttons::P1_START, 0}, //Good
-                        {Buttons::P1_PANEL_UP, 5}, //Good
+                      /*{Buttons::P1_PANEL_UP, 5}, //Good
```
bicarus-dev: we should just delete these

src/spice2x/games/ddr/p4io/p4io.cpp
cchike: When I measured it before, it was something crazy like 2000 kHz. 

I just measured it again and it took 400ms to complete 1,000,000 cycles on average, so yeah 2,500 kHz

bicarus-dev: Kinda feels like the game is calling us as fast as possible on a busy loop... great catch.

src/spice2x/games/ddr/p4io/p4io.cpp
cchike: Done

src/spice2x/games/ddr/p4io/p4io.cpp
cchike: Done

Issue #477, opened 2025-12-29
gitadora arena model cab support


bicarus-dev:
The ones in bold are high on the wish list...

Missing features:

- [x] guitar wail (digital and analog) - #481
- [x] lights? (might be uninteresting if this is all RGB strips...) #492
- [x] volume hook #478 

Network
- [ ] IP address change hook not working - current workaround is to disable all network adapters except one
- [x] local IP check (192.168.x) hook not working (patch exists)

UI issues
- [x] Japanese encoding issue in song select #482
- [x] funky windowed mode scaling issues with overlay - game tries to create dx9 adapter with backbuffer size set to 4k, can we make the backbuffer render at a reasonable resolution like 1080p? addressed by #495
- [ ] windowed mode with 1080p backbuffer?
    - hooking `CreateDeviceEx` and `ResetEx` to force 1080p causes the game to fail to boot.

Graphics issues
- [x] windowed mode toggle without having to modify `app-config-xx.xml` #520
- [ ] **full screen mode working with one 1080p monitor (game wants 4 monitors right now, main monitor must be 4k)**

Overlay
- [x] subscreen overlay in windowed mode #486
- [ ] subscreen overlay in fullscreen for the menu touch screen + mouse support
- [ ] subscreen overlay in fullscreen for the side monitors (maybe, idk, this is a lot of work, but it would be funny)
- [x] windowed options not working (borderless / size adjustment) #494

Audio
- [x] headphone detection i/o? #485 
- [x] is 2ch audio option needed? -> patch exists
- [ ] headphones still don't work even after detection i/o??

Misc
- [ ] game crashes if the main window loses focus during boot

Need verification:
- [x] does the new version still need the processor affinity hack? -> yes, at least on AMD, same issue as before
- [x] does the new cab poll drum i/o more often - if so, midi note algorithm sustain duration needs adjustment #483

_Originally posted by @bicarus-dev in https://github.com/spice2x/spice2x.github.io/issues/476#issuecomment-3694665010_

zxcwdta:
Currently the game does not boot in fullscreen mode, possibly due to the lack of correct monitor configuration.

I am playing around with the `sub_swapchain` thing in `d3d9_device.cpp`. While I am able to stream the small screen to SpiceCompanion, the touch does not work correctly (it maps touch input as cursor movements as supposed to interacting with the small screen)

bicarus-dev:
For the fullscreen issue you should try Only Use One Monitor option.

[...]

PR #481, opened 2025-12-30
gitadora: (arena model) implement wailing, improve picking

bicarus-dev:
#477 

## Description of change
Add analog and digital wail for the arena model guitar.

Fix the picking code to be the same algorithm as extdev - detect rising edges and insert 1 frame of input.

## Testing
Tested x/y/z and digital y up/down. Only Y gets used in game though.

### DIFF
--- a/src/spice2x/games/gitadora/j33i.cpp
+++ b/src/spice2x/games/gitadora/j33i.cpp
@@ -21,6 +21,9 @@ using namespace acioemu;
 // higher but the red dot will reach the edge at +-189
 #define ARENA_MODEL_WAIL_MAX 189
 
+static bool GFDM_GF_PICK_STATE_UP;
+static bool GFDM_GF_PICK_STATE_DOWN;
+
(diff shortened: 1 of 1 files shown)


src/spice2x/games/gitadora/j33i.cpp
```diff
@@ -16,6 +16,14 @@
 
 using namespace acioemu;
 
+// why this number?? no clue, but this is what the test menu shows as the max
+// value in the green box with the red dot; the value display itself can go
+// higher but the red dot will reach the edge at +-189
+#define ARENA_MODEL_WAIL_MAX 189
```
oleg238948234, line 22: that's not the actual limit, i've seen values up to 600 when using a real guitar io unit (and i suspect it can go higher if you shake it real hard)

bicarus-dev: that's fine but from i/o emulation perspective we only map the usable range

PR #483, opened 2025-12-30
gitadora: (arena model) fix j32d encoding & use velocity

oleg238948234:
N/A

## Description of change
in current implementation j32d analog values are not encoded properly, values are supposed to go up to 1023, but they show up as 771 in game

## Testing
tested with M32-007-2025092400, used DM PAD CALIBRATION menu

### DIFF
--- a/src/spice2x/games/gitadora/j32d.cpp
+++ b/src/spice2x/games/gitadora/j32d.cpp
@@ -55,13 +55,15 @@ bool games::gitadora::J32DSerialDevice::parse_msg(
             memset(&payload, 0, sizeof(J32D_IO_STATUS));
 
             auto &buttons = get_buttons();
-            auto high_tom = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumHiTom]) * 1023);
-            auto low_tom = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumLowTom]) * 1023);
-            auto snare = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumSnare]) * 1023);
-            auto floor_tom = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumFloorTom]) * 1023);
-            auto left_cymbal = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumLeftCymbal]) * 1023);
-            auto right_cymbal = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumRightCymbal]) * 1023);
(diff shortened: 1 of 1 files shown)


src/spice2x/games/gitadora/j32d.cpp
```diff
@@ -53,37 +53,25 @@ bool games::gitadora::J32DSerialDevice::parse_msg(
...
-                payload.hihat = 4092;
-            }
+            auto &buttons = get_buttons();
+            auto high_tom = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumHiTom]) * 1023);
+            auto low_tom = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumLowTom]) * 1023);
+            auto snare = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumSnare]) * 1023);
+            auto floor_tom = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumFloorTom]) * 1023);
+            auto left_cymbal = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumLeftCymbal]) * 1023);
+            auto right_cymbal = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumRightCymbal]) * 1023);
+            auto hihat = (uint16_t) (GameAPI::Buttons::getVelocity(RI_MGR, buttons[Buttons::DrumHiHat]) * 1023);
```
bicarus-dev: This doesn't account for HiHatClosed / HiHatHalfOpen; need to check all three and take the max velocity value.

bicarus-dev:
I'm not sure if this is a good idea in general. The MIDI v2_drum algorithm is finely tuned against the digital input polling. I've had multiple people with various drum sets calibrate them.

If we switch to analog (velocity-based), you are skipping over the sustain logic, which would lead to dropped input. 

To elaborate on this... in real life, hits on a drum is instantaneous. It's entirely up to the drum brain / driver to report the velocity value. Some drum brains report the velocity value and keep it for couple milliseconds. Others report it for 1 window message and drop it entirely. For event-based DAW this is not an issue. However, we are poll-based, so if you happen to poll (call getVelocity) between the notes then you can drop notes.

Gitadora (pre-arena model) polls at 60Hz, which is woefully slow, but this means if a drum hit lands within a 16.67ms window then you have a good chance of missing the hit entirely. That's why the sustain duration is 20ms, it's guaranteed to be long enough for one poll.

bicarus-dev:
I see, since we are at device level now it polls more often. It's still poll-based which means we can be inoperable on midi brains that report ON event and immediately OFF message, so the sustain logic is still needed.

src/spice2x/games/gitadora/j32d.cpp
oleg238948234: fixed

PR #484, opened 2025-12-30
gitadora: (arena model) F: drive fix

oleg238948234:
N/A

## Description of change
prevents the game from creating folders on F: drive

## Testing
tested with M32-007-2025092400

### DIFF
--- a/src/spice2x/games/gitadora/gitadora.cpp
+++ b/src/spice2x/games/gitadora/gitadora.cpp
@@ -22,6 +22,7 @@ namespace games::gitadora {
      * Prevent GitaDora from creating folders on F drive
      */
 
+#ifdef SPICE64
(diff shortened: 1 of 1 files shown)


src/spice2x/games/gitadora/gitadora.cpp
```diff
@@ -215,6 +239,10 @@ namespace games::gitadora {
         detour::inline_hook((void *) sys_debug_dip_set_param, libutils::try_proc(
                 system_module, "sys_debug_dip_set_param"));
 
+        detour::iat_try("GetLogicalDrives", GetLogicalDrives_hook, avs::game::DLL_INSTANCE);
```
bicarus-dev: These should be in the `if (aio != nullptr) {` block below

oleg238948234: fixed

PR #518, opened 2026-01-13
Add coin support for SDVX BI2X

russelg:
## Description of change
Add coin handling to SDVX BI2X. I also tried to populate the structs as best I could.

## Testing
Tested via running spice2x on EG final, using F1 to insert coins.


bicarus-dev:
thanks!

PR #574, opened 2026-03-14
cfg: auto light binding, sdvx light formatting, all light test/clear binds

Horocchi:
image

n/a

## Description of change

I've always been annoyed with how long it takes to bind controller lights, made a few improvements

1) added a button+window for automatically matching device lights to game lights with the same name
2) added button(s) to cycle through all bound lights to visually confirm location/function
3) added button to clear all bound lights
4) split SDVX lights into sections (Buttons, Valkyrie, Nemsys, Other) with formatting, sorted based on current game spec

## Testing
Tested with multiple games and controllers (Faucetwo, custom con) and matching always works if the descriptor strings are labeled correctly.

Not every controller labels their LEDs the same was as spice, but for those that do this saves a ton of time clicking every box and testing one by one.

The SDVX table split/formatting is important because myself (and several friends) have tried binding to Wing/Controller lights and wondered why they weren't working in game, when they are legacy Nemsys lights instead of the strip lights handled by Valk/bi2x. I separated the sections visually, with tooltips, plus sort the two sections based on active spec to further minimize the chance of someone trying to use the wrong lights for their game.

### DIFF
--- a/src/spice2x/overlay/windows/config.cpp
+++ b/src/spice2x/overlay/windows/config.cpp
@@ -3054,12 +3054,18 @@ namespace overlay::windows {
                     break;
             }
 
-            // match by name (case insensitive)
+            // match by name (hard match, case insensitive, with P1/P2 prefix fallback)
             std::vector<bool> device_matched(raw_names.size(), false);
+            std::string player_prefix = auto_match_p2 ? "P2 " : "P1 ";
+            auto player_prefix_lower = strtolower(player_prefix);
             for (auto &entry : matched) {
                 auto game_lower = strtolower(entry.game_name);
                 for (int ci = 0; ci < (int) raw_names.size(); ci++) {
-                    if (!device_matched[ci] && game_lower == strtolower(raw_names[ci])) {
+                    if (device_matched[ci]) {
+                        continue;
+                    }
+                    auto dev_lower = strtolower(raw_names[ci]);
+                    if (game_lower == dev_lower || game_lower == player_prefix_lower + dev_lower) {
                         entry.device_name = raw_names[ci];
                         entry.control_index = ci;
                         device_matched[ci] = true;
@@ -3071,7 +3077,6 @@ namespace overlay::windows {
             // soft matching
             bool is_valkyrie_mode = games::sdvx::is_valkyrie_model();
             static const char *RGB[] = {" R", " G", " B"};
-            std::string player_prefix = auto_match_p2 ? "P2 " : "P1 ";
 
             // try to match a game light name, first as-is, then with P1/P2 prefix
             auto try_match = [&](const std::string &game_target, int ci, const char *controller) -> bool {


src/spice2x/overlay/windows/config.cpp
```diff
@@ -2334,30 +2544,135 @@ namespace overlay::windows {
...
+                ImVec4(1.f, 0.7f, 0, 1),
+                "%s", name);
+            if (tooltip) {
+                ImGui::SameLine();
+                ImGui::HelpMarker(tooltip);
+            }
+            ImGui::Separator();
+        };
+
+        if (is_sdvx) {
```
bicarus-dev, line 2583: We'll probably need a better way to categorize lights at API layer (e.g., in sdvx/io.cpp) since not all games have the lights in sequential order, but it can be in a future change.

src/spice2x/overlay/windows/config.cpp
```diff
@@ -2666,6 +2981,328 @@ namespace overlay::windows {
...
+                        raw_names.push_back(
+                            rawinput::SmxDedicabDevice::GetLightNameByIndex(i));
+                    }
+                    break;
+                }
+                default:
+                    break;
+            }
+
+            // match by name
```
bicarus-dev: This works well for single player games, but for 2-player games (IIDX, DDR, etc) almost all controllers are have 1p and 2p as separate controllers and will likely never find a match. e.g., `P1 Button 1` for an input vs. `Button 1` on a controller.

Again, not a blocker for this PR, in the future we should consider other ways of soft-matching things.

src/spice2x/overlay/windows/config.cpp
```diff
@@ -2666,6 +2981,328 @@ namespace overlay::windows {
...
+                ImGuiTableColumnFlags_WidthStretch);
+            ImGui::TableSetupColumn("",
+                ImGuiTableColumnFlags_WidthFixed, 20.f);
+            ImGui::TableSetupColumn("Device Output",
+                ImGuiTableColumnFlags_WidthStretch);
+            ImGui::TableSetupScrollFreeze(0, 1);
+            ImGui::TableHeadersRow();
+
+            for (int i = 0; i < (int) matched.size(); i++) {
+                auto &entry = matched[i];
```
bicarus-dev, line 3387: It seems that this prints all lights that are present in the controller in the end - is that intentional?

bicarus-dev: Looks like this at the end of the table (this is my gen 1 FauceTwo)

Image

src/spice2x/overlay/windows/config.cpp
```diff
@@ -2666,6 +2981,328 @@ namespace overlay::windows {
...
+                default:
+                    break;
+            }
+
+            // match by name
+            std::vector<bool> device_matched(raw_names.size(), false);
+            for (auto &entry : matched) {
+                for (int ci = 0; ci < (int) raw_names.size(); ci++) {
+                    if (!device_matched[ci]
+                            && entry.game_name == raw_names[ci]) {
```
bicarus-dev: Can this be case insensitive match instead? Spice has `Start` but my Faucetwo has `START`...

src/spice2x/overlay/windows/config.cpp
```diff
@@ -2316,15 +2317,224 @@ namespace overlay::windows {
...
+            if (ImGui::IsItemHovered(ImGui::TOOLTIP_FLAGS)) {
+                ImGui::SetTooltip("Unbind all lights.");
+            }
+
+            // test all
+            ImGui::SameLine();
+            ImGui::BeginDisabled(
+                !lights_testing && bound.empty());
+            if (ImGui::Button(test_label)) {
+                if (lights_testing) {
```
bicarus-dev, line 2447: Can we make this run indefinitely until the user presses Stop? It runs a little too fast.

Same for the other test button.

src/spice2x/overlay/windows/config.cpp
```diff
@@ -2666,6 +2981,328 @@ namespace overlay::windows {
...
+                } else {
+                    ImGui::TextDisabled("%s",
+                        entry.game_name.c_str());
+                }
+
+                ImGui::TableNextColumn();
+                if (entry.control_index >= 0) {
+                    ImGui::TextUnformatted("=");
+                } else {
+                    ImGui::TextDisabled("-");
```
bicarus-dev: Should be `!=` or something, not `-`

src/spice2x/overlay/windows/config.cpp
Horocchi: This is intentional, but optional.

I thought it would be useful to be able to see the other controller lights that didn't get bound, such as the different chassis light zones on Faucetwo. You can of course see them in the list on the individual mapping screens but I thought this would give more users who just click Auto and scroll down an indicator that they have more options.

The random null ones are probably something I can add a filter for after testing for them

src/spice2x/overlay/windows/config.cpp
Horocchi: I'll replace these with more relevant symbols

src/spice2x/overlay/windows/config.cpp
Horocchi: The light order is something that could definitely have more global organization, I was planning to go through and re-order/categorize them for different games later and can replace this then

src/spice2x/overlay/windows/config.cpp
Horocchi: Yeah, I was thinking that but didn't know if you'd want an endless loop. I might slow the cycle a bit too

src/spice2x/overlay/windows/config.cpp
Horocchi: Yep, will fix

src/spice2x/overlay/windows/config.cpp
bicarus-dev: Sounds reasonable

src/spice2x/overlay/windows/config.cpp
bicarus-dev: Maybe just blank if one side is empty?

Horocchi:
Will fix these tomorrow, I have a couple ideas I'll try as well

src/spice2x/overlay/windows/config.cpp
Horocchi: Great point, I'll add support for this

Horocchi:
Addressed most of the simple things - see commit message

Working on a more complicated change right now, I'll need your feedback on it so give me a bit

src/spice2x/overlay/windows/config.cpp
```diff
@@ -2315,16 +2319,257 @@ namespace overlay::windows {
...
+            for (int i = 0; i < (int) lights->size(); i++) {
+                if ((*lights)[i].isSet()) {
+                    bound.push_back(i);
+                }
+            }
+            if (lights_test_current >= 0
+                    && lights_test_current
+                        < (int) bound.size()) {
+                auto &cur = (*lights)[
+                    bound[lights_test_current]];
```
bicarus-dev: This is a bit hard to read, we use 100 character columns and sometimes even up to 120. Same for other places in this PR where the line is being wrapped unnecessarily.

src/spice2x/overlay/windows/config.cpp
```diff
@@ -2334,30 +2579,135 @@ namespace overlay::windows {
...
-
-                    // alternatives
-                    int alt_index = 1; // 0 is primary
-                    for (auto &alt : light.getAlternatives()) {
-                        if (alt.isValid()) {
-                            build_light(light, &alt, light_index, alt_index);
-                        }
+        bool is_sdvx = (this->games_selected_name
+            == "Sound Voltex");
+        bool is_valkyrie = is_sdvx
```
bicarus-dev: `games::sdvx::is_valkyrie_model()`

src/spice2x/overlay/windows/config.cpp
Horocchi: Good to know - I'll go through these

src/spice2x/overlay/windows/config.cpp
Horocchi: Would require including `games/sdvx/sdvx.h`, as long as you're fine with that. I checked and it doesn't look like any other games (even IIDX) have game specific lights that require a check so it would just be this I think.

PR #641, opened 2026-04-15
graphics: add option to disable win11 rounded corners on windows

InvoxiPlayGames:
## Description of change
This adds a toggleable option to remove the rounded window corners when running a game, or a subscreen, in a window on Windows 11 and upwards.

## Testing
On Windows 11: Resort Anthem, MÚSECA and EPOLIS were tested and the window corners were removed from both main windows and the TDJ subscreen when running in windowed mode, with no change in fullscreen mode.

On Windows 7: IIDX18 was tested with no change, the call to `DwmSetWindowAttribute` is ignored as `DWMWA_WINDOW_CORNER_PREFERENCE` doesn't exist.

For XP builds, the call is compiled out to avoid linking against dwmapi and windows_dll_compat_checker returns successfully when testing against the XP Embedded image. I can't test it on XP, but it ran on Vista.

### DIFF
--- a/src/spice2x/cfg/screen_resize.h
+++ b/src/spice2x/cfg/screen_resize.h
@@ -60,6 +60,7 @@ namespace cfg {
         //     window = rectangle including the frame
         //     client = just the content area without frames.
         bool window_always_on_top = false;
+        bool window_disable_round_corners = false;
         bool client_keep_aspect_ratio = true;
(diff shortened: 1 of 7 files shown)


bicarus-dev:
I didn't even realize we can opt out. Makes me wonder if we should always opt out... this is perfectly fine though, thank you.

Issue #121, opened 2024-01-11
Support for old card mechanism emulation (IIDX 18 & 19)


HiLordReilo:
Please add support for old card mechanism emulation, because currently there is no way to insert card in Resort Anthem with spicetools.

PR #642, opened 2026-04-15
acio: implement ICCA slotted reader emulation for iidx18

InvoxiPlayGames:
#121

## Description of change
This implements a seperate ICCA card workflow for the slotted readers acting how Resort Anthem expects, enabling cards to be inserted. Additionally this fixes a bug where both card units would be treated the same (for keypad reads) under Resort Anthem.

The slotted reader workflow was implemented as a seperate function for two reasons - it would be nice to not mess with any existing code to avoid breaking all the functional games, and the way the slotted readers are handled by RA seemed different enough to how everything was already designed to where it didn't make sense to keep adding special edge cases for RA.

## Testing
IIDX18 was tested and cards could be inserted on both P1 and P2 side. IIDX19 through 22 was tested to make sure nothing had broken for wavepass emulation.

## Known Issues
In IIDX18, when a card isn't set but the insert key is pressed, after the card timeout the game will error out with an invalid state error.

### DIFF
--- a/src/spice2x/acio/icca/icca.cpp
+++ b/src/spice2x/acio/icca/icca.cpp
@@ -35,24 +35,33 @@ struct ICCA_STATUS_LA9 {
 static_assert(sizeof(struct ICCA_STATUS) == 24, "ICCA_STATUS must be 24 bytes");
 
 enum ICCA_WORKFLOW {
-    STEP,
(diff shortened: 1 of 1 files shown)


bicarus-dev:
Awesome, thanks.

PR #567, opened 2026-03-06
fix various minor issues causing compilation failure with clang

mon:
- Include order and forward decls were breaking button.h because a declared but not defined struct can't be initiated (honestly this file was cooked, crazy include order).
- `MAXINT` is GCC specific, `INT_MAX` is portable.
- InterlockedDecrement takes a LONG, not ULONG
- an imgui printf-style call using a direct string instead of %s (let's ignore the fact that it's actually safe based on how the str is constructed)
- missing `override` on a virtual subclass
- `CALLBACK` on a lambda threw me for a loop, but I believe moving it after the arg list is the correct approach (see if it builds on gcc I guess)
- `std::result_of` was removed in C++20, which the project is built with

PR #660, opened 2026-04-27
Try and preserve the wideness of std::filesystem::path more

mon:
As noted in #567, a filesystem path that contains non-ascii will break a lot if using a clang toolchain.

Luckily, fmtlib has a lossy utf8 convert when you use it to print a path (after including `fmt/std.h`). The vast majority of this diff is just removing `.string()` from paths inside loggings calls.

There are some callsites I _didn't_ touch, mainly the options, because it would be an ABI break to change those to be wide strings and I cbf looking into settings upgrades. There are also some spots (avs mountpath remapping, for example) where the path is guaranteed to be ascii, so I didn't modify them.

ImGui doesn't appear to easily support wide strings (I mean, surely it does, but I'm not gonna look too far into it) so I mostly just left those alone too, with a few spots modified to re-use fmtlib's lossy utf8.

Some of the changes are basically never gonna be hit IRL, like who would put a file with a non-ascii _extension_ along with their modules? But the diff is (I hope) pretty easy to validate as OK.

Testing has been somewhat minimal, I fired up the GCC build of spice2x in a dodgy folder name, got mojibake (running via wine in linux so take that as you will), ran the unmodified clang spice and crashed the same way the reporter did. After modification, I get the exact same mojibake so I assume if the terminal enjoys utf8 it'll display OK.

Claude (only used for review) thinks the commit is fine but is annoyed that I use `fmt::detail` in the appdata censoring, which is part of the private API; personally I don't care because it's pretty stable.

### DIFF
--- a/src/spice2x/api/modules/memory.cpp
+++ b/src/spice2x/api/modules/memory.cpp
@@ -58,7 +58,7 @@ namespace api::modules {
 
         // check if file exists in modules
         if (!fileutils::file_exists(dll_path)) {
-            return error(res, "Couldn't find " + dll_path.string());
+            return error(res, fmt::format("Couldn't find {}", dll_path));
         }
 
         // get module
@@ -118,7 +118,7 @@ namespace api::modules {
 
         // check if file exists in modules
         if (!fileutils::file_exists(dll_path)) {
-            return error(res, "Couldn't find " + dll_path.string());
+            return error(res, fmt::format("Couldn't find {}", dll_path));
         }
 
         // get module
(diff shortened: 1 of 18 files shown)


mon:
One note actually, for the remaining callsites (and actually some of the ones in this PR) we may be able to "just" use u8string instead of string, especially for display (since it'll probably blow up if you try and give a differing codepage to fopen or whatever). Source: https://github.com/mstorsjo/llvm-mingw/issues/534

PR #694, opened 2026-05-13
network: icmp emulation

drmext:
#0

## Description of change
Add `-icmphook` option under spicecfg Development/Network for emulating keepalive ping replies in user mode, so games do not need to open privileged raw ICMP sockets. This makes connecting to networks possible on macOS under Whisky, and on Windows with ICMP firewalled or spice running without Administrator privileges.

## Testing
Tested DDR and SDVX on Windows 10 and macOS Tahoe. Network check status in test menu shows `CONNECTED` when -icmphook is enabled, and `NOT CONNECTED` when it is disabled. Actual network functionality works as expected (under previously impossible conditions).

### DIFF
--- a/src/spice2x/hooks/icmphook_iphlpapi.cpp
+++ b/src/spice2x/hooks/icmphook_iphlpapi.cpp
@@ -5,6 +5,7 @@
 #include <icmpapi.h>
 
 #include <algorithm>
+#include <cstddef>
 #include <cstring>
 
 #include "util/detour.h"
@@ -61,17 +62,15 @@ DWORD WINAPI IcmpSendEcho_hook(
 
     memset(ReplyBuffer, 0, ReplySize);
     auto *reply = reinterpret_cast<ICMP_ECHO_REPLY *>(ReplyBuffer);
+    const size_t data_off = offsetof(ICMP_ECHO_REPLY, Data);
+    const size_t avail = ReplySize > data_off ? ReplySize - data_off : 0;
+    const size_t copy_len = (std::min)(static_cast<size_t>(RequestSize), avail);
     reply->Address = DestinationAddress;
     reply->Status = IP_SUCCESS;
     reply->RoundTripTime = 1;
-    reply->DataSize = RequestSize;
-    const size_t data_off =
-            reinterpret_cast<const uint8_t *>(reply->Data)
-            - reinterpret_cast<const uint8_t *>(reply);
-    const size_t avail = ReplySize > data_off ? ReplySize - data_off : 0;
-    const size_t copy_len = (std::min)(static_cast<size_t>(RequestSize), avail);
+    reply->DataSize = static_cast<USHORT>(copy_len);
     if (copy_len > 0) {
-        memcpy(reply->Data, RequestData, copy_len);
+        memcpy(reinterpret_cast<uint8_t *>(ReplyBuffer) + data_off, RequestData, copy_len);
     }
     SetLastError(ERROR_SUCCESS);
     return 1;
--- a/src/spice2x/hooks/icmphook_iphlpapi.cpp
+++ b/src/spice2x/hooks/icmphook_iphlpapi.cpp
@@ -1,0 +1,103 @@
+#include <winsock2.h>
+#include <windows.h>
+#include <iptypes.h>
+#include <iphlpapi.h>
+#include <icmpapi.h>
+
+#include <algorithm>
+#include <cstring>
+
+#include "util/detour.h"
+#include "util/logging.h"
+
+namespace {
+
+#pragma pack(push, 1)
+struct IcmpHdr {
+    uint8_t type;
+    uint8_t code;
+    uint16_t checksum_be;
+};
+#pragma pack(pop)
+
+DWORD WINAPI (*IcmpSendEcho_orig)(
+        HANDLE IcmpHandle,
+        IPAddr DestinationAddress,
+        LPVOID RequestData,
+        WORD RequestSize,
+        PIP_OPTION_INFORMATION RequestOptions,
+        LPVOID ReplyBuffer,
+        DWORD ReplySize,
+        DWORD Timeout) = nullptr;
+
+DWORD WINAPI IcmpSendEcho_hook(
+        HANDLE IcmpHandle,
+        IPAddr DestinationAddress,
+        LPVOID RequestData,
+        WORD RequestSize,
+        PIP_OPTION_INFORMATION RequestOptions,
+        LPVOID ReplyBuffer,
+        DWORD ReplySize,
+        DWORD Timeout) {
+    if (!IcmpSendEcho_orig) {
+        SetLastError(ERROR_INVALID_FUNCTION);
+        return 0;
+    }
+    DWORD n = IcmpSendEcho_orig(
+            IcmpHandle,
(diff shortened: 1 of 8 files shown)


src/spice2x/hooks/icmphook_net.cpp
```diff
@@ -0,0 +1,647 @@
...
+int WINAPI bind_hook_ws2(SOCKET s, const sockaddr *name, int namelen) {
+    {
+        std::lock_guard<std::recursive_mutex> lock(g_mu);
+        auto it = g_socks.find(s);
+        if (it != g_socks.end()) {
+            if (namelen < (int) sizeof(sockaddr_in)) {
+                WSASetLastError(WSAEINVAL);
+                return SOCKET_ERROR;
+            }
+            auto *in = reinterpret_cast<const sockaddr_in *>(name);
```
bicarus-dev, line 357: should check for bad params here before dereferencing (just call `bind_trampoline_orig` instead in bad cases)

src/spice2x/hooks/icmphook_net.cpp
```diff
@@ -0,0 +1,647 @@
...
+
+int WINAPI setsockopt_hook(
+        SOCKET s, int level, int optname, const char *optval, int optlen) {
+    {
+        std::lock_guard<std::recursive_mutex> lock(g_mu);
+        auto it = g_socks.find(s);
+        if (it != g_socks.end()) {
+            EmuSock *es = it->second.get();
+            if (level == SOL_SOCKET && optlen >= (int) sizeof(DWORD)) {
+                auto v = reinterpret_cast<const DWORD *>(optval);
```
bicarus-dev, line 593: same here, should check for bad params just in case

src/spice2x/hooks/icmphook_net.cpp
```diff
@@ -0,0 +1,647 @@
...
+            WSASetLastError(WSAEWOULDBLOCK);
+            return SOCKET_ERROR;
+        }
+    } else if (es->rx_timeout_ms == 0) {
+        if (es->queue.empty()) {
+            WSASetLastError(WSAETIMEDOUT);
+            return SOCKET_ERROR;
+        }
+    } else if (es->rx_timeout_ms == INFINITE) {
+        es->cv.wait(lock, wait_pred);
```
bicarus-dev: Can this race with `release_icmp_socket` - while we wait, `release_icmp_socket` can call `notify_all` and then immediately delete the pointer, but this routine can continue to work on it.

src/spice2x/hooks/icmphook_iphlpapi.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+#pragma pack(push, 1)
+struct IcmpHdr {
+    uint8_t type;
+    uint8_t code;
+    uint16_t checksum_be;
+};
+#pragma pack(pop)
+
+DWORD WINAPI (*IcmpSendEcho_orig)(
```
bicarus-dev: suggest:

`decltype(IcmpSendEcho) *IcmpSendEcho_orig = nullptr`

src/spice2x/hooks/icmphook_iphlpapi.cpp
```diff
@@ -0,0 +1,95 @@
...
+    if (ih->type != 8 || ih->code != 0) {
+        return 0;
+    }
+
+    memset(ReplyBuffer, 0, ReplySize);
+    auto *reply = reinterpret_cast<ICMP_ECHO_REPLY *>(ReplyBuffer);
+    reply->Address = DestinationAddress;
+    reply->Status = IP_SUCCESS;
+    reply->RoundTripTime = 1;
+    reply->DataSize = RequestSize;
```
bicarus-dev: should this be `copy_len`?

src/spice2x/hooks/icmphook_iphlpapi.cpp
```diff
@@ -0,0 +1,95 @@
...
+    const size_t data_off =
+            reinterpret_cast<const uint8_t *>(reply->Data)
```
bicarus-dev: the math seems off here, ` reinterpret_cast<const uint8_t *>(reply->Data)` is always zero;  are you trying to find the offset of Data field within the struct?

src/spice2x/hooks/icmphook_net.cpp
```diff
@@ -0,0 +1,664 @@
...
+        return j == g_socks.end() || !j->second->queue.empty();
+    };
+
+    EmuSock *es = it->second.get();
+    if (es->nonblocking) {
+        if (es->queue.empty()) {
+            WSASetLastError(WSAEWOULDBLOCK);
+            return SOCKET_ERROR;
+        }
+    } else if (es->rx_timeout_ms == 0) {
```
bicarus-dev: apparently `SO_RCVTIMEO` set to 0 is supposed to be indefinite wait in blocking case, so the same as `INFINITE`

src/spice2x/hooks/icmphook_net.cpp
```diff
@@ -0,0 +1,660 @@
...
+    }
+    if (lpOverlapped != nullptr) {
+        WSASetLastError(WSAEOPNOTSUPP);
+        return SOCKET_ERROR;
+    }
+    if (dwBufferCount != 1 || !lpBuffers || !lpNumberOfBytesRecvd) {
+        WSASetLastError(WSAEINVAL);
+        return SOCKET_ERROR;
+    }
+    char stack_buf[65536];
```
bicarus-dev: this is a huge buffer on stack memory; since we have no idea how deep into the stack we are when we get called, this might overflow in some extreme cases

I think we can call `recvfrom_hook` and copy directly onto `lpBuffers[0].buf` without intermediate copy

src/spice2x/hooks/icmphook_net.cpp
```diff
@@ -0,0 +1,660 @@
...
+    }
+    return closesocket_orig(s);
+}
+
+int WINAPI bind_hook_ws2(SOCKET s, const sockaddr *name, int namelen) {
+    std::unique_lock<std::recursive_mutex> lock(g_mu);
+    auto it = g_socks.find(s);
+    if (it != g_socks.end()) {
+        if (!name || namelen < (int) sizeof(sockaddr_in)) {
+            lock.unlock();
```
bicarus-dev: style wise, we have a lot of places where we use a RAII lock but you are unnecessarily calling .unlock() in some paths (but not others) - would recommend just not manually unlocking.

ditto for other functions (search for `lock.unlock()`)

src/spice2x/hooks/icmphook_net.cpp
```diff
@@ -0,0 +1,660 @@
...
+        uint32_t bind_pref_host_order,
+        bool has_bind) {
+    if (peer_host_order == k_loopback_host) {
+        return k_loopback_host;
+    }
+    if (has_bind && bind_pref_host_order != 0) {
+        return bind_pref_host_order;
+    }
+    ULONG buf_len = 0;
+    GetAdaptersInfo(nullptr, &buf_len);
```
bicarus-dev: we should just do this once and cache the results

bicarus-dev:
I think we are almost there, thanks for your patience.

PR #698, opened 2026-05-22
Automated PIN Macro after auto-inserting a card

Biduleman:
## Description of change
Adds a way to automatically use the PIN Macro for P1, P2 or both when Auto Card Insert is used.

### Launcher changes: 
 - Added a checkbox to enable/disable the feature
 - Added a textbox for each player to set what to look for in the logs.

### How it works:
The feature requires Auto Card Insert and PIN Macro to be enabled and set.
It uses the Auto Card Insert value to enable the feature for P1, P2 or both, with a "master" checkbox to completely disable the feature. It then relies on the current PIN Macro code to send the PIN to the games.

The games logs send a message when a new scene is loaded with the name of the scene. By hooking on the logs, we can see when a PIN is required and send the relevant macro. Games have different scene names, and sometimes one for each player, so the field is configurable per game per player. These could be hard-coded in a lookup table but I didn't want to recompile the project every time I wanted to test, and I don't have all the games to get this info from.

The PIN Macro thread is used to hook to the logs and check if the defined string appears, which triggers the PIN macro just like a button press would.

Setting nothing for the log trigger disables the feature for that player. It's useful if someone wants to have Auto Card Insert enabled for both players but only Auto Pin on 1 player, but is mainly there so the check won't happen if nothing is set.

## Testing
I used SDVX and DDR to have a single and a two player test with the feature both enabled and disabled. I also tested multiple combinations of PIN Macro settings and Auto Card Insert settings with no apparent issues. The PIN is automatically entered as soon as the game requests it.

## Discussion

I know that hooking to the logs might not be the best solution. I'm very open to discussions about this, and understand that this might be a deal breaker for this feature. I made this to use with my setup and thought I'd try opening this PR to see if the feature would be useful for others.

I haven't found another way to make sure we're on the correct screen to launch the PIN Macro, and sending the macro every time the Auto Card Insert finishes had sync issues where the end of the PIN would actually be registered by the game.

It would also be much more convenient for the users to hardcode a lookup table for every games' log-trigger but I don't have the games currently so I can't check the logs to see what would be best.


src/spice2x/launcher/options.cpp
```diff
@@ -2464,6 +2464,56 @@ static const std::vector<OptionDefinition> OPTION_DEFINITIONS = {
...
+            "auto-trigger for P1 in this game. Only effective when Auto PIN Macro is "
+            "enabled.",
+        .type = OptionType::Text,
+        .setting_name = "",
+        .category = "Network",
+    },
+    {
+        // spice2x_AutoPinMacroTrigger1
+        .title = "Auto PIN Macro Trigger Log String (P2)",
+        .name = "sp2x-autopinmacrotrigger1",
```
bicarus-dev: remove all of this and simplify to `.name = "autopinmacrotrigger1`

sp2x prefix is no longer being used, so aliases aren't needed either

same for the other one

src/spice2x/launcher/options.h
```diff
@@ -253,6 +253,9 @@ namespace launcher {
             IIDXSubMonitorOverride,
             spice2x_IIDXEmulateSubscreenKeypadTouch,
             spice2x_AutoCard,
+            spice2x_AutoPinMacro,
```
bicarus-dev: remove spice2x_ prefix

src/spice2x/launcher/options.cpp
```diff
@@ -2464,6 +2464,56 @@ static const std::vector<OptionDefinition> OPTION_DEFINITIONS = {
             {"both", ""},
         },
     },
+    {
```
bicarus-dev, line 2467: you could probably just get rid of this boolean option and enable the feature automatically if Trigger0 or Trigger1 options are set for each player

src/spice2x/misc/eamuse.cpp
```diff
@@ -44,13 +45,31 @@ static uint8_t AUTO_INSERT_CARD_CACHED_DATA[2][8];
...
 
+static bool pin_macro_log_hook(void *, const std::string &data, logger::Style, std::string &) {
+    for (int unit = 0; unit < 2; unit++) {
+        if (AUTO_PIN_MACRO_TRIGGER[unit].empty()) {
+            continue;
+        }
+        if (data.find(AUTO_PIN_MACRO_TRIGGER[unit]) == std::string::npos) {
+            continue;
+        }
+        if (AUTO_INSERT_CARD[unit] && !PIN_MACRO_VALUES[unit].empty()) {
```
bicarus-dev: these checks should be done once upfront (when you check to see if logger hook should be added)

src/spice2x/launcher/options.cpp
Biduleman: Done

src/spice2x/launcher/options.h
Biduleman: Done

src/spice2x/launcher/options.cpp
Biduleman: Good call! It's done.

src/spice2x/misc/eamuse.cpp
Biduleman: I have moved the checks what were done against constants outside of the hook and now pass the array of yes/no flags to the hook.

src/spice2x/launcher/options.cpp
```diff
@@ -2464,6 +2464,34 @@ static const std::vector<OptionDefinition> OPTION_DEFINITIONS = {
...
+        .name = "autopinmacrotrigger0",
+        .desc =
+            "Substring matched against game log output to detect when Player 1's PIN entry "
+            "screen is ready. When the substring appears in any log line, the PIN macro is "
+            "typed for Player 1 only (Auto Card Insert must also be enabled for P1). "
+            "Configure per game (each game tab in spicecfg keeps its own value); for "
+            "example, Sound Voltex emits 'ARD_ENTRY_AUTH_SCENE'. Leave blank to disable "
+            "auto-trigger for P1 in this game.",
+        .type = OptionType::Text,
+        .category = "Network",
```
bicarus-dev: `Network (Advanced)`, so that these are next to `PIN Macro` options

src/spice2x/launcher/options.cpp
```diff
@@ -2464,6 +2464,34 @@ static const std::vector<OptionDefinition> OPTION_DEFINITIONS = {
...
     },
+    {
+        // AutoPinMacroTrigger0
+        .title = "Auto PIN Macro Trigger Log String (P1)",
+        .name = "autopinmacrotrigger0",
+        .desc =
+            "Substring matched against game log output to detect when Player 1's PIN entry "
+            "screen is ready. When the substring appears in any log line, the PIN macro is "
+            "typed for Player 1 only (Auto Card Insert must also be enabled for P1). "
+            "Configure per game (each game tab in spicecfg keeps its own value); for "
```
bicarus-dev: Remove `Configure per game` - this is implied.

src/spice2x/misc/eamuse.cpp
```diff
@@ -345,15 +377,17 @@ void eamuse_pin_macro_start_thread() {
...
                 for (int unit = 0; unit < 2; unit++) {
                     if (PIN_MACRO_VALUES[unit].empty()) {
                         continue;
                     }
-                    if (overlay_buttons &&
+                    bool key_press = overlay_buttons &&
                         (!overlay::OVERLAY || overlay::OVERLAY->hotkeys_triggered()) &&
-                        GameAPI::Buttons::getState(RI_MGR, overlay_buttons->at(PIN_MACRO_TRIGGER_KEYS[unit]))) {
+                        GameAPI::Buttons::getState(RI_MGR, overlay_buttons->at(PIN_MACRO_TRIGGER_KEYS[unit]));
+                    bool auto_request = AUTO_PIN_MACRO_REQUEST[unit].exchange(false);
```
bicarus-dev, line 392: For debug purposes, can we log something here when auto_request is detected?

src/spice2x/misc/eamuse.cpp
```diff
@@ -324,6 +340,22 @@ void eamuse_pin_macro_start_thread() {
...
+            !AUTO_PIN_MACRO_TRIGGER[unit].empty() &&
+            AUTO_INSERT_CARD[unit] &&
+            !PIN_MACRO_VALUES[unit].empty();
+    }
+
+    // register scene log hook so the macro fires on the per-game trigger string,
+    // but only if at least one player is eligible
+    if (AUTO_PIN_MACRO_PLAYER_ACTIVE[0] || AUTO_PIN_MACRO_PLAYER_ACTIVE[1]) {
+        logger::hook_add(pin_macro_log_hook, nullptr);
+    }
```
bicarus-dev, line 360: log info here to indicate that feature is being enabled

src/spice2x/misc/eamuse.cpp
```diff
@@ -324,6 +340,22 @@ void eamuse_pin_macro_start_thread() {
...
     PIN_MACRO_THREAD_ACTIVE = true;
 
+    // a unit is eligible for auto-trigger only if all the static prerequisites are
+    // satisfied at startup; precomputing this lets the log hook skip per-line checks
+    // on values that never change at runtime
+    for (int unit = 0; unit < 2; unit++) {
+        AUTO_PIN_MACRO_PLAYER_ACTIVE[unit] =
+            !AUTO_PIN_MACRO_TRIGGER[unit].empty() &&
+            AUTO_INSERT_CARD[unit] &&
+            !PIN_MACRO_VALUES[unit].empty();
```
bicarus-dev, line 349: might be worth logging a user configuration error; e.g., if `AutoPinMacroTrigger0` is enabled but `pinmacro0` is not

src/spice2x/misc/eamuse.cpp
```diff
@@ -324,6 +340,22 @@ void eamuse_pin_macro_start_thread() {
     // set active
     PIN_MACRO_THREAD_ACTIVE = true;
 
+    // a unit is eligible for auto-trigger only if all the static prerequisites are
+    // satisfied at startup; precomputing this lets the log hook skip per-line checks
+    // on values that never change at runtime
+    for (int unit = 0; unit < 2; unit++) {
+        AUTO_PIN_MACRO_PLAYER_ACTIVE[unit] =
+            !AUTO_PIN_MACRO_TRIGGER[unit].empty() &&
+            AUTO_INSERT_CARD[unit] &&
```
bicarus-dev: auto insert card should not be a requirement, I should be able to manually insert card and have the pin entered automatically

bicarus-dev (changes requested):

src/spice2x/misc/eamuse.cpp
Biduleman: Done

src/spice2x/misc/eamuse.cpp
Biduleman: Done

src/spice2x/misc/eamuse.cpp
Biduleman: Done

src/spice2x/launcher/options.cpp
Biduleman: Done

src/spice2x/launcher/options.cpp
Biduleman: Done

src/spice2x/misc/eamuse.cpp
Biduleman: Done

PR #705, opened 2026-05-27
[+] Hide sides on gitadora 𝛿

hykilpikonna:
## Description of change
Added an option to only disable the left/right screens of gitadora gw delta while exposing the touch panel screen.

## Testing
Tested by hand on CachyOS + hyprland + gamescope + wine + vnc + phone + cat 🐈‍⬛ 

image

### DIFF
--- a/src/spice2x/hooks/graphics/graphics.cpp
+++ b/src/spice2x/hooks/graphics/graphics.cpp
@@ -359,8 +359,8 @@ static HWND WINAPI CreateWindowExA_hook(DWORD dwExStyle, LPCSTR lpClassName, LPC
     bool is_sdvx_sub_window = avs::game::is_model("KFC") && window_name.ends_with(" Sub Screen");
     bool is_popn_sub_window = avs::game::is_model("M39") && window_name.ends_with("Sub Screen");
     bool is_gfdm_sub_window = games::gitadora::is_arena_model() && window_name.ends_with("SMALL");
-    bool is_gfdm_left_window = games::gitadora::is_arena_model() && window_name == "LEFT";
-    bool is_gfdm_right_window = games::gitadora::is_arena_model() && window_name == "RIGHT";
+    bool is_gfdm_left_window = games::gitadora::is_arena_model() && window_name.ends_with("LEFT");
+    bool is_gfdm_right_window = games::gitadora::is_arena_model() && window_name.ends_with("RIGHT");
 
     // update style / ex-style
     if (is_tdj_sub_window || is_sdvx_sub_window || is_gfdm_sub_window || is_popn_sub_window) {
--- a/src/spice2x/hooks/graphics/graphics.cpp
+++ b/src/spice2x/hooks/graphics/graphics.cpp
@@ -39,6 +39,8 @@ struct CaptureData {
 HWND TDJ_SUBSCREEN_WINDOW = nullptr;
 HWND SDVX_SUBSCREEN_WINDOW = nullptr;
 HWND GFDM_SUBSCREEN_WINDOW = nullptr;
+static HWND GFDM_LEFT_WINDOW = nullptr;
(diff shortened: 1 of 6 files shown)


bicarus-dev:
Please disregard my previous comments - I realized that you're only fixing windowed mode, not fullscreen.

As the option is written today, it's difficult to understand (from the option name) that this only affects windowed mode. Maybe we need to have a dedicated enum option for selecting the layout in windowed mode instead of two booleans.

src/spice2x/hooks/graphics/graphics.cpp
```diff
@@ -356,6 +359,8 @@ static HWND WINAPI CreateWindowExA_hook(DWORD dwExStyle, LPCSTR lpClassName, LPC
     bool is_sdvx_sub_window = avs::game::is_model("KFC") && window_name.ends_with(" Sub Screen");
     bool is_popn_sub_window = avs::game::is_model("M39") && window_name.ends_with("Sub Screen");
     bool is_gfdm_sub_window = games::gitadora::is_arena_model() && window_name.ends_with("SMALL");
+    bool is_gfdm_left_window = games::gitadora::is_arena_model() && window_name == "LEFT";
```
bicarus-dev: stick to `ends_with` as most games tend to change the title every version

src/spice2x/launcher/options.cpp
```diff
@@ -1089,6 +1089,18 @@ static const std::vector<OptionDefinition> OPTION_DEFINITIONS = {
...
     },
+    {
+        // GitaDoraArenaHideSideWindows
+        .title = "GitaDora Arena Hide Side Screens (EXPERIMENTAL)",
+        .name = "gdahideside",
+        .desc = "For Arena Model windowed mode: hide the LEFT and RIGHT side screens, "
+            "but keep the dedicated SMALL touch screen window.\n\n"
+            "This is independent from GitaDora Arena Disable Subscreens. If both are "
+            "enabled, Disable Subscreens takes priority and SMALL is also hidden.",
+        .type = OptionType::Bool,
```
bicarus-dev: make this an `Enum` and allow for 1-window mode / 2-window mode / 4 window mode (default).

bicarus-dev: and rename the option accordingly to reflect above

hykilpikonna:
> Please disregard my previous comments - I realized that you're only fixing windowed mode, not fullscreen.
> 
> As the option is written today, it's difficult to understand (from the option name) that this only affects windowed mode. Maybe we need to have a dedicated enum option for selecting the layout in windowed mode instead of two booleans.

ok yea it's meant to be used with windowed mode

> make this an `Enum` and allow for 1-window mode / 2-window mode / 4 window mode (default).

Is this what you have in mind?

Before:

```
  nothing              = fullscreen, 4 monitors
  -w                   = 4 windows
  -w -gdaonewindow     = 1 window
  -gdaonewindow        = fullscreen single-monitor mode
```

After:

```
  nothing              = fullscreen, 4 monitors
  -gdaonewindow        = fullscreen single-monitor mode
  -w                   = 4 windows
  -w -gdawindows=4     = 4 windows
  -w -gdawindows=2     = 2 windows: main + SMALL
  -w -gdawindows=1     = 1 window: main only
  -w -gdaonewindow     = same as above
  -gdawindows=1/2/4    = ignore/warn? or should we force -w when -gdawindows exist?
```

bicarus-dev:
Looks good. Regarding:

`  -gdawindows=1/2/4    = ignore/warn? or should we force -w when -gdawindows exist?`

I think we should just ignore and log a warning message.


and just for completeness,

`  nothing              = fullscreen, 4 monitors`
this defaults to single-monitor mode if we detect fewer than 4 active (real) monitors.

hykilpikonna:
Should be good now

PR #707, opened 2026-05-28
overlay: implement dx11 backend

### LINKED ISSUE #134: DX11 hook compatibility and support for Spice Overlay
Unity games like CCJ and QKS use DX11, but since Spice Overlay only supports DX9, it does not function properly.

bicarus-dev:
Fixes #134 

## Description of change
Add ImGui DX11 backend implementation, and integrate into the spice overlay.

The tricky part is that some of the Unity games:

1. Launch multiple windows, and
2. Can be resized (even full screen games launch at a certain resolution and then expand to fit desktop resolution)
3. lazy load DX11 DLLs

This PR also adds screenshot functionality, but screen resize is not implemented.

## Testing

Seems to be working for most games. Need to do final tests for:

- [x] CCJ (attract movie plays as well)
- [x] Busou Shinki (title movie plays)
- [x] MFG
- [x] QuizKnock
- [x] Polaris Chord
- [x] check dx9 for regression

### DIFF
--- a/src/spice2x/CMakeLists.txt
+++ b/src/spice2x/CMakeLists.txt
@@ -517,6 +517,11 @@ set(SOURCE_FILES ${SOURCE_FILES}
         hooks/graphics/backends/d3d9/d3d9_fake_swapchain.cpp
         hooks/graphics/backends/d3d9/d3d9_swapchain.cpp
         hooks/graphics/backends/d3d9/d3d9_texture.cpp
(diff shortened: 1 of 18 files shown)


aixxe:
Just a heads up that this breaks compatibility with Windows 7, at least with the stock cabinet image.

image

bicarus-dev:
😭

PR #709, opened 2026-05-29
overlay: prevent save button shifting patches text

drmext:
#0

## Description of change
When opening the patches tab by pressing F4 in game, the Save button doesn't appear until after initially selecting/deselecting a patch, then it shifts all of the patch text below. To solve this annoyance, this keeps the Save button in a fixed position (starts greyed out until a selection) and no longer shifts the text.

## Testing
Confirmed saving selected patches still works properly in the overlay and standalone cfg


bicarus-dev:
Thanks!

PR #716, opened 2026-05-29
Add GITADORA Arena windowed monitor layout support, allows the SMALL  touch window to be resized

9DYnastY:
N/A

## Description of change
Adds GITADORA Arena windowed monitor layout options for the GITADORA, LEFT, RIGHT, and SMALL windows.

This also applies the selected window border style to all four GITADORA Arena windows, and allows the SMALL touch window to be resized in  windowed mode while keeping touch coordinates updated.

## Testing
- Manually tested GITADORA Arena windowed mode:
  - default, resizable, and borderless window styles
  - per-window monitor placement
  - SMALL resize
  - SMALL touch coordinate mapping after resize


src/spice2x/hooks/graphics/graphics_windowed.cpp
```diff
@@ -51,6 +55,173 @@ static const DWORD SETWINDOWPOS_NOOP =
...
+    }
+    return cfg::WindowDecorationMode::Default;
+}
+
+bool graphics_gitadora_is_borderless_windowed() {
+    return GRAPHICS_WINDOWED &&
+        graphics_effective_window_decoration() == cfg::WindowDecorationMode::Borderless;
+}
+
+static const std::optional<std::string> &graphics_gitadora_monitor_for_window_name(
```
bicarus-dev: instead of `std::optional<std:string>` I would just use `std::string` and treat empty strings as invalid (which you're already doing anyway)

src/spice2x/hooks/graphics/graphics.cpp
```diff
@@ -701,6 +943,25 @@ static BOOL WINAPI SetWindowPos_hook(HWND hWnd, HWND hWndInsertAfter,
         return TRUE;
     }
 
+    if (gitadora_should_apply_window_placement(hWnd)) {
+        const auto window_name = gitadora_window_name_for_hwnd(hWnd);
+        int width = cx;
+        int height = cy;
+        if (window_name != nullptr &&
+            graphics_gitadora_apply_window_monitor(
```
bicarus-dev: gitadora versions, delta included, are known to soft lock during scene transitions once the game is resized; did you not encounter any of that here?

src/spice2x/hooks/graphics/graphics.h
```diff
@@ -56,6 +56,10 @@ extern bool GRAPHICS_WINDOW_ALWAYS_ON_TOP;
 extern bool GRAPHICS_WINDOW_BACKBUFFER_SCALE;
 extern bool GRAPHICS_WINDOW_DISABLE_ROUNDED_CORNERS;
 extern std::optional<HWND> GRAPHICS_HOOKED_WINDOW;
+extern std::optional<std::string> GRAPHICS_GITADORA_MAIN_MONITOR;
```
bicarus-dev: should just be `std::string` as well and treat empty strings as "not set"

src/spice2x/launcher/options.cpp
```diff
@@ -2307,6 +2307,55 @@ static const std::vector<OptionDefinition> OPTION_DEFINITIONS = {
         .type = OptionType::Bool,
         .category = "Graphics (Windowed)",
     },
+    {
+        // GitaDoraWindowedMainMonitor
+        .title = "GitaDora Windowed GITADORA Monitor",
```
bicarus-dev: `Main Monitor`

src/spice2x/hooks/graphics/graphics.cpp
9DYnastY: Thanks, updated.

I did not encounter the soft lock behavior in my manual testing, but I agree the runtime MoveWindow/SetWindowPos enforcement was risky, so I changed the implementation.

Monitor placement is now applied at window creation. Later game-side MoveWindow/SetWindowPos placement calls for configured GITADORA windows are suppressed instead of resizing again during scene transitions.

I also replaced the monitor override optionals with strings and renamed the main monitor option title to "GitaDora Windowed Main Monitor".

src/spice2x/hooks/graphics/graphics.cpp
```diff
@@ -39,6 +39,7 @@ struct CaptureData {
 
 HWND TDJ_SUBSCREEN_WINDOW = nullptr;
 HWND SDVX_SUBSCREEN_WINDOW = nullptr;
+static HWND GFDM_MAIN_WINDOW = nullptr;
```
bicarus-dev: I don't think we need this at all if we have GRAPHICS_HOOKED_WINDOW

9DYnastY: Updated.👌

PR #720, opened 2026-05-30
cfg: d3d9 standalone configurator

drmext:
## Description of change
**Configurator (spicecfg.exe)**

- Attempts to create a D3D9 device on startup and uses the hardware ImGui DX9 path when successful; falls back to the existing software rasterizer automatically if D3D9 init fails.
- Rewrites the configurator window loop: refresh-rate-aware render timer, minimize/resize/`WM_DISPLAYCHANGE` handling, direct Win32 -> ImGui input (keyboard, mouse, wheel), and optimized software painting via `SetDIBitsToDevice` instead of per-frame GDI `HBITMAP` allocation.

**Overlay / input (scoped to standalone configurator)**

- Skips expensive per-frame rawinput polling in `ImGui_ImplSpice_NewFrame` when `CONFIGURATOR_STANDALONE` is set (Win32 messages drive input instead).
- Adds software-renderer idle-frame detection (`sw_pixels_dirty` + draw-data hash) so spicecfg only repaints when pixels actually change.
- Adds `ImGuiWindowFlags_NoScrollWithMouse` on the configurator root window to prevent scroll jolt on non-scrollable child widgets.

### DIFF
--- a/src/spice2x/cfg/configurator_wnd.cpp
+++ b/src/spice2x/cfg/configurator_wnd.cpp
@@ -254,9 +254,9 @@ void cfg::ConfiguratorWindow::run() {
     SetFocus(this->hWnd);
 
     // match the render timer to the monitor refresh rate so scrolling stays smooth
-    // on 60/120/144 Hz panels. Idle CPU is still bounded by overlay::sw_pixels_dirty
-    // (software path) and D3DPRESENT_INTERVAL_ONE vsync (D3D9 path); the timer is
-    // also paused entirely when the window is minimized (see WM_SIZE handler).
+    // on 60/120/144 Hz panels. Idle CPU is bounded by overlay::sw_pixels_dirty
+    // (software path) and overlay::d3d9_frame_dirty (D3D9 path); the timer is also
+    // paused entirely when the window is minimized (see WM_SIZE handler).
     const UINT hz = detect_monitor_refresh_hz(this->hWnd);
     this->timer_interval_ms = std::max<UINT>(1, 1000 / hz);
     log_info("configurator", "render timer {} ms ({} Hz)",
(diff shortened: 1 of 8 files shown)


bicarus-dev:
Insane. Haven't looked at it yet but

>direct Win32 -> ImGui input (keyboard, mouse, wheel)

this will finally fix touch in spicecfg, at least on the right path to fixing it

bicarus-dev:
Seems like `spice -cfg` `spice64 -cfg` still work which is great..

Do you think it would be worthwhile to add a debug option that forces software rendering just in case?

drmext:
Maybe we could put spicecfg-sw.exe in extras

src/spice2x/cfg/configurator_wnd.cpp
```diff
@@ -110,62 +369,167 @@ LRESULT CALLBACK cfg::ConfiguratorWindow::window_proc(HWND hWnd, UINT uMsg, WPAR
...
+                // BeginScene/EndScene, and Present(). Avoids any CPU rasterization.
+                state->device->Clear(0, nullptr, D3DCLEAR_TARGET,
+                    D3DCOLOR_RGBA(20, 18, 18, 255), 1.0f, 0);
+                if (SUCCEEDED(state->device->BeginScene())) {
+                    if (overlay::OVERLAY) {
+                        overlay::OVERLAY->render();
+                    }
+                    state->device->EndScene();
+                }
+                HRESULT hr = state->device->Present(nullptr, nullptr, nullptr, nullptr);
```
bicarus-dev: Claude says checking for `D3DERR_DEVICELOST` is not sufficient and can cause the app to lock up; e.g., when you do ctrl+L to lock the screen and then unlock (and I can confirm this does happen)

and the suggested fix is

```cpp
HRESULT cl = device->TestCooperativeLevel();
if (cl == D3DERR_DEVICENOTRESET) {
    overlay->reset_invalidate();
    if (SUCCEEDED(device->Reset(&pp))) overlay->reset_recreate();
}
// else if cl == D3DERR_DEVICELOST: just skip this frame and try again next tick
```

bicarus-dev:
> Maybe we could put spicecfg-sw.exe in extras

I'd like to avoid this if possible because it'll add another couple minutes to the build time

PR #724, opened 2026-05-31
[O] Filter GetIpAddrTable to selected network adapter

hykilpikonna:
## Description of change
Filter GetIpAddrTable to selected network adapter. 

Fixes the IP address error (5-1506-0000) on gitadora gw delta when multiple network adapters are present.

## Testing
Tested on Gitadora GW Delta with both WiFi and Tailscale adapters present.


bicarus-dev:
Awesome, thanks.

Issue #73, opened 2023-11-16
Make the card ID more correct


gar6a9e:
For old e-amusement pass (NFC Type V), the correct starting mID is `E00401`, not `E004`.

````patch
Index: overlay/windows/card_manager.h
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/overlay/windows/card_manager.h b/overlay/windows/card_manager.h
--- a/overlay/windows/card_manager.h	(revision fb92150af8aa7cc896a2e61b8f287355a70e8c5f)
+++ b/overlay/windows/card_manager.h	(date 1700155746718)
@@ -6,7 +6,7 @@
 
     struct CardEntry {
         std::string name = "unnamed";
-        std::string id = "E004000000000000";
+        std::string id = "E004010000000000";
         bool selected = false;
     };
 
Index: launcher/options.cpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/launcher/options.cpp b/launcher/options.cpp
--- a/launcher/options.cpp	(revision fb92150af8aa7cc896a2e61b8f287355a70e8c5f)
+++ b/launcher/options.cpp	(date 1700155746729)
@@ -85,7 +85,7 @@
         .name = "card0",
         .desc = "Set a card number for reader 1. Overrides the selected card file.",
         .type = OptionType::Text,
-        .setting_name = "E004000000000000",
+        .setting_name = "E004010000000000",
         .category = "Network",
         .sensitive = true,
     },
@@ -94,7 +94,7 @@
         .name = "card1",
         .desc = "Set a card number for reader 2. Overrides the selected card file.",
         .type = OptionType::Text,
-        .setting_name = "E004000000000000",
+        .setting_name = "E004010000000000",
         .category = "Network",
         .sensitive = true,
     },
Index: README.md
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/README.md b/README.md
--- a/README.md	(revision fb92150af8aa7cc896a2e61b8f287355a70e8c5f)
+++ b/README.md	(date 1700155746726)
@@ -95,7 +95,7 @@
   "id": 1,
   "module": "card",
   "function": "insert",
-  "params": [0, "E004000000000000"]
[...]

PR #748, opened 2026-06-10
improve E004 card generation and placeholder pcbid

drmext:
#73

## Description of change
Generated cards now start with E0040100 to fully match real cards. The placeholder example `E004010000000000` is replaced with `E0040100FFFFFFFF`.

The invalid placeholder/fallback PCBID `04040000000000000000` is replaced with `01201000000000010101`, which contains a valid header/checksum, making it closer resemble a real PCBID while still clearly being an example.


bicarus-dev:
Regarding the fallback PCBID - if I'm remembering things correctly it was actually intentional to generate a PCBID value that can be passed to offline servers (like asphyxia) that do not care to check valid PCBIDs but fails checks on properly implemented servers. 

I don't feel strongly about this though to block the PR but just a thought... in reality actual servers will require a *registered* PCBID so the whole point is moot.

drmext:
Hmm, this made me think of another concern with cards. The proper format has about 256x fewer combinations (4.3 billion vs 1.1 trillion), which raises the chance of a collision with a real card (although still astronomically low to have someone with the real card not be able to register it on a private network because someone else already generated the exact same one). So it could be argued that we need to specifically *not* generate E0040100 so there's no chance of a collision by a real card.

bicarus-dev:
I'm not too worried about conflicting with real cards.

Is there a real benefit to making this more accurate (using E0040100) - servers today don't seem to care.

drmext:
No, I think the only benefit is aesthetics of correctness. Same goes for the valid header/checksum PCBID fallback, because older games that actually require it aren't supported in spice.

Issue #746, opened 2026-06-10
capture.get_jpg blocks during game load, causing Spice Companion Mirror to disconnect


jessemarthin:
## Game and version
IIDX (Lightning)

## Version of spice2x
Latest main from github.com/spice2x/spice2x.github.io

## Describe the issue
When using stock Spice Companion's Mirror feature, the view disconnects and returns
to KeyPad during game loading.
Expected: Mirror stays connected (or at least the API keeps responding).

Actual: Companion stops receiving JPEG frames. The API call `capture.get_jpg` appears
to block in `graphics_capture_receive_jpeg()` when D3D9 capture cannot complete during
loading (no Present / capture failure paths).
I have a fix ready (2s timeout, skip signalling on D3D9 failure, last-frame JPEG cache)
and will open a PR if this issue is accepted.

## Summary
When using Spice Companion's Mirror feature, the connection drops back to TouchPad
during game loading.
## Root cause (suspected)
`capture.get_jpg` waits indefinitely in `graphics_capture_receive_jpeg()` when D3D9
screen capture cannot complete (e.g. no Present calls during loading). The API
thread blocks and Companion times out.

## Expected behavior
Mirror should stay connected during loading, or at least not block the API thread.

## Actual behavior
Mirror disconnects and returns to KeyPad during loading.

## Notes
I have a fix prepared as a PR (timeout + skip signalling + last-frame JPEG cache).

bicarus-dev:
Thanks. Which Companion app are you using? The iOS client is completely different from the others.

jessemarthin:
> Thanks. Which Companion app are you using? The iOS client is completely different from the others.

Thanks for the clarification.
I am using the stock iOS Spice Companion app (unmodified, from the App Store).
I have only tested this on the iOS client so far. I have not tested Android or any other Companion client.
On iOS, the issue shows up in the Mirror feature: during game loading, Mirror disconnects and returns to the KeyPad view. From the spice2x side, this appears to correlate with `capture.get_jpg` blocking in `graphics_capture_receive_jpeg()` when D3D9 capture cannot complete during loading.
I understand the iOS client implementation differs from the others. I mentioned Companion mainly as the user-visible symptom. The proposed fix is entirely on the spice2x side (capture timeout, skip signalling on D3D9 failure, and last-frame JPEG fallback), and I verified that stock iOS Companion keeps Mirror connected with that spice2x-side patch.

bicarus-dev:
[...]

PR #750, opened 2026-06-11
Fix capture.get_jpg blocking during game load for Companion Mirror

jessemarthin:
Add a timeout and skip signalling to the D3D9 capture path so the API thread no longer waits indefinitely when Present stops during loading. Return the last successful JPEG frame as a fallback.

#746

## Description of change
* Fix `capture.get_jpg` blocking indefinitely when D3D9 screen capture cannot
  complete during game loading (e.g. no Present calls).
* Add a 2-second timeout to `graphics_capture_receive_jpeg()` and cancel
  pending capture requests on timeout.
* Signal capture skip from D3D9 failure paths via `graphics_capture_skip()`.
* Cache the last successful JPEG per screen and return it as a fallback
  when a new capture fails.
Tested with **stock iOS Spice Companion** only. I have not tested Android or
other Companion clients. The iOS client implementation differs from the
others; this fix is server-side only.
No config file, CLI option, or API schema changes.

## Testing
* Built locally with llvm-mingw cross-compiler (WSL).
* Docker build (`src/spice2x/build_docker.sh`): Pending
* Tested with stock iOS Spice Companion on iPad: Mirror remains connected
  during game loading instead of returning to KeyPad.

### DIFF
--- a/src/spice2x/hooks/graphics/graphics.cpp
+++ b/src/spice2x/hooks/graphics/graphics.cpp
@@ -5,6 +5,7 @@
 
 #include "graphics.h"
 
+#include <chrono>
 #include <set>
 #include <vector>
 #include <mutex>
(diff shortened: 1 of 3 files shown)


src/spice2x/hooks/graphics/graphics.cpp
bicarus-dev: this `empty()` check probably needs to be under the lock now?

jessemarthin: Good catch, thanks. You're right — `graphics_capture_cancel_pending()` now removes
entries from the API thread, so the `empty()` check could race and `back()` /
`pop_back()` could run on an empty deque. Moved the check under the lock.

PR #870, opened 2026-08-17
graphics: rewrite screenshot and api capture image processing

bicarus-dev:
#0

## Description of change
Significantly speeds up API screen capture and D3D9 screenshots saving. Two reasons for doing this:

1. We now have a 4K game (GITADORA) and existing capture code was taking multiple seconds.
2. Renewed user interest on streaming as we have a couple more companion apps in active development.

**API screen capture (streaming), 1280x720:** 14.3ms -> 6.3ms per frame. Back buffer copies go to pooled `D3DPOOL_SYSTEMMEM` surfaces via `GetRenderTargetData` instead of allocating a lockable render target every frame, and TooJpeg is replaced with libjpeg-turbo (encode 9.8ms -> 3.0ms). MSAA remains unsupported

**Screenshots for GITADORA arena model, across 4 screens with one of them 4K**: 4068ms -> 124ms. `D3DXSaveSurfaceToFileA` is replaced with fpng (encode 4043ms -> 76ms) and the screens encode in parallel. Dropping D3DX also removes the `d3dx9_43.dll` ... `d3dx9_24.dll` probing loop, so screenshots no longer fail outright on machines with no D3DX9 runtime installed.

Screenshot surfaces are read on the present thread, so no D3D call reaches another thread for screenshots. This fixes a hang in DDR X2 introduced earlier in the branch: its device has no internal locking, and reading the surface on a pool thread while the present thread sat inside `GetRenderTargetData` left the game's own render thread deadlocked.

## Testing

- **GITADORA** (arena model, D3D9Ex, 4K main plus three subscreens, windowed) with
  `-screenshotsub`: three sets of four screenshots, images verified correct. Completion
  order differs between sets, so the screens really are encoding in parallel.
- **LovePlus** (KLP, plain D3D9, 768x1360): covers the inline path used by games whose
  image processing must not leave the present thread. 
- **API screen capture** through a companion app: live video correct throughout.
- **Print Screen** bound as the screenshot key: the clipboard copy succeeded on every shot.
- Quitting the game after capturing leaves no `IDirect3DDevice9` reference count warning,
  so the pooled readback surfaces are released along with the device.


LupinThidr:
No more copilot man
cl

PR #891, opened 2026-08-24
api: add card lookup endpoint

Avimitin:

N/A

## Description of change

I am developing [spice.nimabe.net](https://spice.nimabe.net), a static frontend for SpiceAPI. It already supports generating and inserting new card IDs, but SpiceAPI does not provide a way to discover cards already configured on the game machine.

This makes the frontend inconvenient for existing users who generated and saved their card IDs locally. They currently have to open each card file on the game machine and manually copy its ID into the frontend.

This change adds a `card.get_cards` API function that enumerates the current cards for readers supported by the running game. Each entry contains:

- reader index
- canonical 16-character card ID
- `source`, set to `file` or `override`
- `file_name` for file-backed cards

File-backed cards expose only the save-file basename. Active `-card0` and `-card1` values are returned with `source: "override"` and no `file_name`, instead of representing an override as a file. spicefe derives the useful default import names `card0` and `card1` from the reader index. Reader enumeration uses `eamuse_get_game_keypads()` so games with only one reader do not expose a second card.

This allows spicefe and other API clients to offer one-click import for existing cards, without requiring users to manually find and copy their card IDs.

Because card IDs are sensitive, `card.get_cards` can only be used when the operator has configured an API password. No full filesystem path is exposed. Missing, unreadable, and invalid cards are omitted.

The existing `card.insert` function remains unchanged. To avoid requiring a password for the whole card module, this change adds support for password requirements on individual API functions while preserving existing module-wide password behavior.

Access to configured card paths and runtime card overrides is synchronized because API requests may read them concurrently with configuration updates.

The Python and Dart API wrappers and README documentation have been updated for the new function and source metadata.

## Testing

- Ran the complete `src/spice2x/build_docker.sh` build successfully.
- Built all required 32-bit and 64-bit targets.
- Built the Windows XP-compatible 32-bit targets.
- Passed static import checks.
- Passed Windows 7 and Windows XP compatibility checks.
- Completed release packaging successfully.
- Verified both updated Dart wrappers are correctly formatted.
- Ran all 106 spicefe tests successfully.
- `git diff --check` passes.

*This implementation was prepared with assistance from OpenAI Codex.*

### DIFF
--- a/src/spice2x/README.md
+++ b/src/spice2x/README.md
@@ -144,6 +144,13 @@ restriction is that the ID has to be a valid 64-bit unsigned integer.
     `busy`
 
 #### Card
+- get_cards()
+  - returns the current card ID and source for each active card reader
+  - each entry contains `index`, `card_id`, and `source`
+  - `source` is `file` or `override`
+  - file entries also contain `file_name`, without the full configured path
+  - `-card0` and `-card1` overrides do not contain `file_name`
+  - this function only works when an API password is configured
 - insert(index: uint, card_id: hex)
   - inserts a card which gets read by the emulated card readers for the game
   - index has to be either 0 (for P1) or 1 (for P2)
--- a/src/spice2x/api/controller.cpp
+++ b/src/spice2x/api/controller.cpp
@@ -337,9 +337,11 @@ bool Controller::process_request(ClientState *state, const char *in, size_t in_s
             if (module->name == request.module) {
                 module_found = true;
 
-                // check password force
-                if (module->password_force && this->password.empty() && request.function != "session_refresh") {
-                    Value err("Module requires the password to be set.");
+                // check password requirement
+                if (module->requires_password(request.function)
+                        && this->password.empty()
+                        && request.function != "session_refresh") {
+                    Value err("Function requires the password to be set.");
                     response.add_error(err);
                     break;
                 }
--- a/src/spice2x/api/modules/card.cpp
+++ b/src/spice2x/api/modules/card.cpp
@@ -1,4 +1,5 @@
 #include "card.h"
+#include <fstream>
 #include <functional>
 #include "external/rapidjson/document.h"
 #include "util/logging.h"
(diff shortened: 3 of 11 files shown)


bicarus-dev:
Does this include the card IDs provided via -card0 and -card1 args?

src/spice2x/api/controller.cpp
```diff
@@ -338,7 +338,9 @@ bool Controller::process_request(ClientState *state, const char *in, size_t in_s
                 module_found = true;
 
                 // check password force
-                if (module->password_force && this->password.empty() && request.function != "session_refresh") {
+                if (module->requires_password(request.function)
+                        && this->password.empty()
+                        && request.function != "session_refresh") {
                     Value err("Module requires the password to be set.");
```
bicarus-dev: Error text probably needs update since it's no longer at module level - maybe `function requires the password...`

src/spice2x/api/modules/card.cpp
```diff
@@ -11,8 +12,50 @@ using namespace rapidjson;
...
+        require_password("get");
+    }
+
+    /**
+     * get()
+     */
+    void Card::get(Request &req, Response &res) {
+        auto &alloc = res.doc()->GetAllocator();
+
+        for (size_t index = 0; index < 2; index++) {
```
bicarus-dev: instead of assuming 2 card readers, check against `eamuse_get_game_keypads`

src/spice2x/api/modules/card.cpp
```diff
@@ -11,8 +12,50 @@ using namespace rapidjson;
...
     Card::Card() : Module("card") {
+        functions["get"] = std::bind(&Card::get, this, _1, _2);
         functions["insert"] = std::bind(&Card::insert, this, _1, _2);
+        require_password("get");
+    }
+
+    /**
+     * get()
+     */
+    void Card::get(Request &req, Response &res) {
```
bicarus-dev: `get_cards` since `get` just sounds like you're getting the current card as opposed to enumerating all cards

Avimitin:
> Does this include the card IDs provided via -card0 and -card1 args?

Now included, command-line override will now returns correspond default filename card0 or card1. Changes are squashed and fixup. All requested changes are applied.

src/spice2x/README.md
```diff
@@ -144,6 +144,12 @@ restriction is that the ID has to be a valid 64-bit unsigned integer.
     `busy`
 
 #### Card
+- get_cards()
+  - returns the current card ID and source name for each active card reader
+  - each entry contains `index`, `card_id`, and `file_name`
+  - only the file name is returned, not the full configured path
+  - command line card ID overrides are returned with `card0` or `card1` as the file name
```
bicarus-dev: maybe a `source` field would help to distinguish between where the card comes from (`override`, `file`, etc) instead of trying to reuse the file_name field

Avimitin: Implemented in latest commit.

PR #905, opened 2026-09-06
Give options a designated initializer to make crossreferencing their usage easier

mon:
## Description of change
This always bothered me, you have a commandline arg and it's a massive pain in the ass to work out where it's used because the flag never really matches the enum name.

Array designated initializers are a C99 extension that clang+gcc both support. I suspect msvc still blows up, but are people still using that? I can change it to a std::invoke lambda instead.

Chose to drop the (unused?) option sorting func because, well, unused.

## Testing
Builds in CI, didn't do much more intense testing.

### DIFF
--- a/src/spice2x/launcher/options.cpp
+++ b/src/spice2x/launcher/options.cpp
@@ -91,10 +91,10 @@ bool launcher::USE_CMD_OVERRIDE = false;
 
 /*
  * Option Definitions
- * Be aware that the order must be the same as in the enum launcher::Options!
  */
-static const std::vector<OptionDefinition> OPTION_DEFINITIONS = {
-    {
+static const std::vector<OptionDefinition> OPTION_DEFINITIONS = std::invoke([]() {
+    std::vector<OptionDefinition> c{launcher::Options::_Count};
+    c[launcher::Options::GameExecutable] = {
         .title = "Override Game Executable & Disable Hooks",
         .name = "exec",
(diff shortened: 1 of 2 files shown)


bicarus-dev:
Can't break MSVC, I've seen plenty of people using that.

mon:
Shall I add a CI step for MSVC then? Then I could move to the other approach with confidence

bicarus-dev:
That would be appreciated of course! Apparently it'll produce a billion warnings though.

People occasionally submit fixes to make MSVC work again so the current state may not even compile.

mon:
Can't be that many people on MSVC because it appears to have been broken for the last 17 months :P CI is so very slow, PR coming assuming it succeeds.

mon:
So, this now works but I was considering yet another useless refactor: ditch the enum, turn the thing into a packed struct, and expose that for people who want to iterate it as a `std::span`. The static init function is absolutely massive, if it were turned into a `constexpr` it'd be quite pleasant.

That would require a larger change, that being changing the overlay code to accept string_view instead of just const char*, but I think it's doable since imgui itself uses start/end of a string and does not need the NUL terminator.

bicarus-dev:
Sounds good to me. This is one of the worst parts in this code base so any improvement is welcome

mon:
Changed my mind; the options parsing *also* creates a vector where order is important and indexed via the enum. This is fine for now.

PR #907, opened 2026-09-07
CI: add MSVC build and fix building with MSVC

mon:
Yess yessss let's continue supporting the worst compiler toolchain :smiling_imp:

Fell out of the changes in #905

CI changes were claude because I cannot be arsed with yaml.

Can explain the individual code fixes if ya want any more context.


mon:
Also yeah, MSVC has been broken _at least_ since the github repo got source back in March last year. Personally I would want to just drop support for it entirely because we're in The Future and mingw and docker ain't so bad, but the changes aren't that bad I guess.

bicarus-dev:
Yeah.. I have mixed feeling about it so I'm not sure. 

On one hand a lot of people still only know how to use MSVC and this keeps the door open for them. On the other hand I know things are just straight up broken (linker can be weird and put modules into wrong VA leading to games like Gitadora not loading at all)

mon:
In that case (and considering nobody else noticed MSVC was broken in over a year), perhaps it's better to eject MSVC considering it introduces extra bugs of its own?

Could potentially introduce a new build.ps1/build.bat for windows that uses either the official llvm-mingw toolchain on windows or my hacked up XP version, which is a whole lot easier to understand than docker.

bicarus-dev:
Is Docker that hard for people? The idea was to have people just install Docker and run the batch script without setting up anything..

I'm only aware of one MSVC bug which is that it (at least in the past) did not respect some of the linker flags that asks spice64 to be placed in lower VA, since not doing that causes Gitadora to crash. Other than that it's probably still useful for a lot of development, even if I'll never ship a release using it.

I think of it the opposite way - how many people gave up on building spice because it doesn't build in MSVC, and just walked away without submitting a fix...

mon:
Yeah that's fair. In which case, keep it compatible

bicarus-dev:
Thanks for helping with this!

